Issue #35, opened 2017-10-23
New Matcher: .toBeArray()


mattphillips:
**Feature Request**

### .toBeArray()

Use `.toBeArray` when checking if a value is an `Array`.

```js
test('passes when value is an array', () => {
  expect([]).toBeArray();
  expect([1]).toBeArray();
  expect(true).not.toBeArray();
});
```

stofolus:
I can work on this

mattphillips:
all yours @stofolus

PR #50, opened 2017-10-23
Added toBeArray

stofolus:
### What
feature - Added toBeArray matcher

### Why
#35 

### Housekeeping

- [X] Unit tests
- [X] Add yourself to contributors list (`yarn contributor`)

### DIFF
--- a/src/matchers/toBeArray/__snapshots__/index.test.js.snap
+++ b/src/matchers/toBeArray/__snapshots__/index.test.js.snap
@@ -1,0 +1,17 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`.not.toBeArray fails when given an array 1`] = `
+"<dim>expect(</><red>received</><dim>).not.toBeArray(</><dim>)</>
+
+Expected value to not be a string received:
+  <red>[]</>"
+`;
+
+exports[`.toBeArray fails when not given an array 1`] = `
+"<dim>expect(</><red>received</><dim>).toBeArray(</><dim>)</>
+
+Expected value to be a string:
+  <green>true</>
+Received:
+  <red>false</>"
+`;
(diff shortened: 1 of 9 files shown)


src/matchers/toBeArray/predicate.test.js
```diff
@@ -0,0 +1,12 @@
+import each from 'jest-each';
+import predicate from './predicate';
+
+describe('toBeArray Predicate', () => {
+  it('returns true when given a string', () => {
```
mattphillips: I think the test description is meant to be: 

`returns true when given an array`

src/matchers/toBeArray/index.js
```diff
@@ -0,0 +1,28 @@
...
+const passMessage = received => () =>
+  matcherHint('.not.toBeArray', 'received', '') +
+  '\n\n' +
+  'Expected value to not be an array received:\n' +
+  `  ${printReceived(received)}`;
+
+const failMessage = received => () =>
+  matcherHint('.toBeArray', 'received', '') +
+  '\n\n' +
+  'Expected value to be an array:\n' +
```
mattphillips: This message should be like the passMessage:
For example:
```js
const failMessage = received => () =>
  matcherHint('.toBeArray', 'received', '') +
  '\n\n' +
  'Expected value to be an array received:\n' +
  `  ${printReceived(received)}`;
```

src/matchers/toBeArray/__snapshots__/index.test.js.snap
```diff
@@ -0,0 +1,17 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`.not.toBeArray fails when given an array 1`] = `
+"<dim>expect(</><red>received</><dim>).not.toBeArray(</><dim>)</>
+
+Expected value to not be a string received:
```
mattphillips: I think the tests are failing because of these snapshots.

If you run the tests with `yarn test:watch` you should see that the snapshots are different. You can then update it by pressing `u` while watching the tests.

You can then check the snapshot it should say `Expected value to not be an array received:` instead of `Expected value to not be a string received:`

stofolus:
Everything should be fixed now

mattphillips:
Good job @stofolus thanks a lot

Issue #36, opened 2017-10-23
New Matcher: .toBeString()


mattphillips:
**Feature Request**

### .toBeString()

Use `.toBeString` when checking if a value is a `String`.

```js
test('passes when value is a string', () => {
  expect('').toBeString();
  expect('hello').toBeString();
  expect(true).not.toBeString();
});
```

stofolus:
I can work on this

PR #51, opened 2017-10-23
To be string

olsio:
### What
Added the toBeString matcher.

### Why
Issue [#36](https://github.com/mattphillips/jest-extended/issues/36)

### Notes
I saw the previous pull request to late but this PR also added the case to match String objects created with new String('gewgew')


mattphillips:
Good job @olsio thanks for your help :smile: don't forget to star the repo :star:

Issue #39, opened 2017-10-23
New Matcher: .toStartWith(prefix)


mattphillips:
**Feature Request**

### .toStartWith(prefix)

Use `.toStartWith` when checking if a `String` starts with a given `String` prefix.

```js
test('passes when value is starts with given string', () => {
  expect('hello world').toStartWith('hello');
  expect('hello world').not.toStartWith('world');
});
```

GaryLeutheuser:
Hi, I've started work on this.

mattphillips:
All yours 🥇

PR #53, opened 2017-10-23
Add to start with matcher

GaryLeutheuser:
### What
Implementation of the `toStartWith` matcher.

### Why
This implementation addresses #39.

### Housekeeping

- [x] Unit tests

### DIFF
--- a/src/matchers/toStartWith/index.js
+++ b/src/matchers/toStartWith/index.js
@@ -5,7 +5,9 @@ import predicate from './predicate';
 const passMessage = (prefix, string) => () =>
   matcherHint('.not.toStartWith', 'received', '') +
   '\n\n' +
-  'Expected string to not start with received:\n' +
+  'Expected string to not start with:\n' +
+  `  ${printReceived(prefix)}\n` +
+  'Received:\n' +
   `  ${printReceived(prefix)}`;
 
 const failMessage = (prefix, string) => () =>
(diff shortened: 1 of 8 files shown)


GaryLeutheuser:
I've just noticed that I haven't added an `export` to `src/matchers/index.js`, I'll do that now.

src/matchers/toStartWith/predicate.js
```diff
@@ -0,0 +1 @@
+export default (prefix, string) => prefix === string.substring(0, prefix.length);
```
mattphillips: You can also use `string.startsWith(prefix)` here. Your solution is fine too :smile:

src/matchers/toStartWith/index.js
```diff
@@ -0,0 +1,28 @@
+import { matcherHint, printExpected, printReceived } from 'jest-matcher-utils';
+
+import predicate from './predicate';
+
+const passMessage = (prefix, string) => () =>
```
mattphillips: I think the error message would be better to show the expected too.

```js
const passMessage = (prefix, string) => () =>
  matcherHint('.not.toStartWith') +
  '\n\n' +
  'Expected string to not start with:\n' +
  `  ${printExpected(string)}`;
  'Received:\n' +
  `  ${printReceived(prefix)}`;
```

mattphillips:
Couple of minor points. Great job! Don't forget to star and share the project :smile:

src/matchers/toStartWith/index.js
GaryLeutheuser: Definitely - it's been added.

src/matchers/toStartWith/predicate.js
GaryLeutheuser: Wow, that's really nice. I switched it over.

mattphillips:
Nice one @GaryLeutheuser 🎉

PR #69, opened 2017-10-24
Add toContainEntry matcher

### LINKED ISSUE #17: New Matcher:  .toContainEntry([key, value])
**Feature Request**

### .toContainEntry([key, value])

Use `.toContainEntry` when checking if an object contains the provided entry.

```js
test('passes when object contains given entry', () => {
  const o = { a: 'foo', b: 'bar', c: 'baz' };
  expect(o).toContainEntry(['a', 'foo']);
  expect(o).toContainEntry(['b', 'bar']);
  expect(o).toContainEntry(['c', 'baz']);
  expect(o).not.toContainEntry(['a', 'qux']);
});
```

danielr18:
### What
Add toContainEntry matcher

### Why
Closes #17 

### Housekeeping

- [x] Unit tests


src/matchers/toContainEntry/predicate.js
```diff
@@ -0,0 +1 @@
+export default (obj, [key, value]) => obj.hasOwnProperty && obj.hasOwnProperty(key) && obj[key] === value;
```
mattphillips: I think this could be extended out to do a deep equals i.e. 

```js
import { equals } from 'expect/build/jasmine_utils';

export default (obj, [key, value]) => obj.hasOwnProperty && obj.hasOwnProperty(key) && equals(obj[key], value);
```

danielr18: Right, that sounds good, change should also be done in the other containEntry matchers to keep it consistent

mattphillips:
Thanks @danielr18 I'll add the deep check gives me a chance to code ha 😆

PR #75, opened 2017-10-24
Implemented toBeEmpty

olsio:
I added lodash dependency which will be useful to almost all matchers.

I also updated the other available dependencies.

### What
https://github.com/mattphillips/jest-extended/issues/9

### Housekeeping

- [ ] Unit tests

### DIFF
--- a/src/matchers/toBeEmpty/predicate.js
+++ b/src/matchers/toBeEmpty/predicate.js
@@ -1,0 +1,3 @@
+import _ from 'lodash';
+
+export default value => _.isEmpty(value);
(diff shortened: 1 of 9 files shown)


src/matchers/toBeEmpty/predicate.js
```diff
@@ -0,0 +1,3 @@
+import _ from 'lodash';
+
+export default value => _.isEmpty(value);
```
mattphillips: Lodash's `isEmpty` seems to deem more as empty than this matcher should `_.isEmpty(1) // true`

The matcher should only consider:
`Use .toBeEmpty when checking if a String '', Array [] or Object {} is empty.`

I think for now it might be worth leaving lodash out

mattphillips:
Thanks @olsio 👍

PR #111, opened 2018-01-21
Add toBeArrayOfSize

jsdevtom:
### What
toBeArrayOfSize

### Why
I (and I should imagine many others) need it in my(/their) project(s) to migrate from https://github.com/JamieMason/Jasmine-Matchers

### Notes
I used a slightly different syntax in the message functions. Please confirm that this is OK.

### Housekeeping

- [x] Unit tests
- [X] Add yourself to contributors list (`yarn contributor`)


src/matchers/toBeArrayOfSize/predicate.js
```diff
@@ -0,0 +1,8 @@
+import toBeArray from '../toBeArray/predicate';
+
+/**
```
jsdevtom: Is JSDoc OK here?

src/matchers/toBeArrayOfSize/index.test.js
```diff
@@ -0,0 +1,47 @@
...
+
+describe('.toBeArrayOfSize', () => {
+  {
+    const size = 1;
+    test(`passes when given an array of size 1, where size === ${size}`, () => {
+      expect([1]).toBeArrayOfSize(size);
+    });
+  }
+  {
+    const size = 134;
```
jsdevtom: What do you think to these scoped variables? They have been use to avoid writing errors.

yarn.lock
```diff
@@ -2,6 +2,14 @@
 # yarn lockfile v1
```
jsdevtom: Should I be committing the yarn.lock file?

README.md
```diff
@@ -210,6 +210,18 @@ test('passes when value is an array', () => {
 });
 ```
 
+#### .toBeArrayOfSize()
```
mattphillips: Could you add a link to this in the table of contents at the top of the readme?

src/matchers/toBeArrayOfSize/index.js
```diff
@@ -0,0 +1,34 @@
+import { matcherHint, printExpected, printReceived } from 'jest-matcher-utils';
+
+import predicate from './predicate';
+
+const passMessage = (actual, expected) => () => `${matcherHint('.not.toBeArrayOfSize', 'received', '')}
```
mattphillips: If you change the matcher hint to just use the name of the matcher (negated in this case) like:
```js
matcherHint('.not.toBeArrayOfSize')
```

Under the hood this will default to `matcherHint('.not.toBeArrayOfSize', 'received', 'expected')` which wraps the expected section of the print out in a green color.

src/matchers/toBeArrayOfSize/index.js
```diff
@@ -0,0 +1,34 @@
...
+
+Expected value to not be an array of size:
+
+  ${printExpected(expected)}
+
+Received:
+
+  ${printReceived(actual.length)}`;
+
+const failMessage = (actual, expected) => () => `${matcherHint('.toBeArrayOfSize', 'received', '')}
```
mattphillips: Same as above comment matcherHint can be: `matcherHint('.toBeArrayOfSize')`

src/matchers/toBeArrayOfSize/index.js
```diff
@@ -0,0 +1,34 @@
+import { matcherHint, printExpected, printReceived } from 'jest-matcher-utils';
+
+import predicate from './predicate';
+
+const passMessage = (actual, expected) => () => `${matcherHint('.not.toBeArrayOfSize', 'received', '')}
+
+Expected value to not be an array of size:
```
mattphillips: I think we should be consistent with newlines and follow the pattern used in other matchers: 

```
Expected value to not be an array of size:
  ${printExpected(expected)}
Received:
  ${printReceived(actual.length)}
```

src/matchers/toBeArrayOfSize/index.test.js
mattphillips: I'm not torn either way 😃

src/matchers/toBeArrayOfSize/predicate.js
mattphillips: Generally I don't think we have used it elsewhere but I'm OK with it if you want to leave it 👍

src/matchers/toBeArrayOfSize/predicate.js
```diff
@@ -0,0 +1,8 @@
...
+ * @param actual
+ * @param size
+ * @returns {boolean}
+ */
+export default (actual, size) => toBeArray(actual) && actual.length === size;
```
mattphillips: I think there is a test missing for when the actual is not an array

yarn.lock
mattphillips: I'd say probably not in this case as none of the dependencies have changed

mattphillips:
Hey @jsdevtom thanks for submitting this, I've added a couple of tiny comments but overall it looks very good 😄 

I wasn't aware of `Jasmine-Matchers` - Do you think there are other popular matchers that `jest-extended` could benefit from?

src/matchers/toBeArrayOfSize/index.js
jsdevtom: Can you please confirm if this is the correct exact string literal for this example so that I can amend the others?

```js
const passMessage = (actual, expected) => () => `Expected value to not be an array of size:
  ${printExpected(expected)}
Received:
  ${printReceived(actual.length)}`;
```

mattphillips: Looks good to me 👍

jsdevtom:
I think there are lots of great ones! I couldn't name any specific ones though

src/utils/index.js
```diff
@@ -4,4 +4,8 @@ export const contains = (list, value) => {
   return list.findIndex(item => equals(item, value)) > -1;
 };
 
+export const determinePropertyMessage = (actual, property, message = 'Not Accessible') => {
```
jsdevtom: I couldn't inline this logic because jest didn't recognize the coverage which I provided by testing the error message itself so I thought that I would generalize it and test it that way. What do you think?

jsdevtom:
Is there something wrong with the packages? Or is it my code? https://travis-ci.org/jest-community/jest-extended/builds/336928854?utm_source=github_status&utm_medium=notification

mattphillips:
@jsdevtom don't worry about the tests failing in node 6 on travis - it's not your code :smile: there seems to be an issue in Jest (I think) related to this issue https://github.com/facebook/jest/issues/5431

mattphillips:
All LGTM @jsdevtom thanks for sending this PR dude! 👍

mattphillips:
@jsdevtom I've published this in v0.6.0 🎉

jsdevtom:
Awesome. No problem, thank you for the guidance!

Issue #114, opened 2018-03-22
A matcher that checks that two arrays contain the same elements


denizdogan:
**Feature Request**

Description: A matcher that checks that two arrays contain the same elements. Examples:

```js
expect([1, 2]).toMatchArray([2, 1]) // order doesn't matter!
expect([1, 2, 2]).toMatchArray([2, 1, 2]) // same number of 2s, all good
expect([1, 2]).not.toMatchArray([1, 2, 2]) // one extra 2, FAIL
expect([1, 2, 2]).not.toMatchArray([1, 2]) // one extra 2, FAIL
```

Possible solution: Given arrays A and B, check that they are equal in length. For each element in A, check that the element is equal to any element in B. If it is, remove the element from both arrays. Loop until both arrays are empty.

mattphillips:
Hey @denizdogan 

I like the sounds of this matcher! I'm not too sure about the name though, as `toMatchArray` sounds like it would need to be identical in terms of order.

How about something like `.toIncludeValues([values])`?

`jest-extended` currently has similar APIs in [`.toIncludeAllMembers([members])`](https://github.com/jest-community/jest-extended/tree/master/src/matchers/toIncludeAllMembers) and [`.toIncludeAnyMembers([members])`](https://github.com/jest-community/jest-extended/tree/master/src/matchers/toIncludeAnyMembers) which can be used for reference in terms of structuring the matcher.

How would you feel about sending a PR for this?

denizdogan:
@mattphillips Yeah, I'm not sure what to call it, `.toIncludeValues` isn't 100% accurate either, it sounds like that's the same as `.toIncludeAllMembers` to me! :)

Maybe `.toEqualMembers`? It's hard to describe the matcher concisely. Sure, I can make a PR for this!

mattphillips:
@denizdogan sorry missed the notification on this! I say lets just start a PR and we can decide on a name after 😆

denizdogan:
@mattphillips Yeah, I'll start working on the PR today or tomorrow, I just had a lot of things to take care of IRL 👍

mattphillips:
merged in #122

AyobamiH:
I just started learning jest and I have been bashing my head against the keyboard before finding this, I still suffer from not being able to

le0-0:
[...]

PR #122, opened 2018-04-18
Add toMatchArray matcher

denizdogan:
### What

Feature

### Why

Because this matcher was needed for a project of mine :)

### Notes

As discussed in #114 the name could probably be improved. I found it rather difficult to formulate a concise and correct description of the matcher in README.md, help would be appreciated.

I haven't used Jest's snapshot feature much, so it's possible that the snapshot file could be improved. I'm honestly not sure what it's supposed to contain exactly.

### Housekeeping

- [x] Unit tests


denizdogan:
I'm not sure what the CI build failure is all about, it occurred a few times while I was developing this patch locally. I haven't touched `toHaveBeenCalledBefore`, so I don't think it's my fault! ;)

xprn:
I think that the reason CI fails is because the functions called are executed too quickly in between each other.

mattphillips:
Thanks @denizdogan all looks good. I'm still a little unsure of the name, I think I will rename it to be closer to the other `Array` matchers as `.toIncludeSameMembers`

Issue #117, opened 2018-04-11
Date API Proposal


thiagodp:
**Feature Request**

Description:
API for Date matchers. Some matchers are inspired by [Jasmine Matchers](https://github.com/JamieMason/Jasmine-Matchers).

Possible solution:

Jasmine-like matchers:
```javascript
expect(date).toBeDate(); // Matcher added
expect(date).toBeValidDate();  // Matcher added

expect(date).toBeAfter(otherDate); // Matcher added
expect(date).toBeBefore(otherDate); // Matcher added

expect(object).toHaveDate(memberName);
expect(object).toHaveDateAfter(memberName, date);
expect(object).toHaveDateBefore(memberName, date);
```
Additional matchers:
```javascript
expect(date).toBeAfterOrEqualTo(otherDate); // or toBeGreaterThanOrEqualTo
expect(date).toBeBeforeOrEqualTo(otherDate); // or toBeLessThanOrEqualTo
expect(date).toBeBetween(startDate, endDate);
expect(date).toDiffInDays(otherDate, days); // e.g., expect( '12/25/2018' ).toDiffInDays( '12/24/2018', 1 );

expect( date ).toHaveDay( day );
expect( date ).toHaveMonth( month );
expect( date ).toHaveYear( year );
```

mattphillips:
Nice one @thiagodp I like these a lot! Fancy sending a PR for any of them? :smile:

thiagodp:
Hello @mattphillips ,
added #118 with `toBeDate()`. It's a first step, but I'll try to add more soon.

CoreyKovalik:
I believe the following shouldn't pass, since 0 value unix codes are common bugs

This currently passes, but I believe should fail:
```
expect(new Date(0)).toBeValidDate()
```

thiagodp:
Hi @CoreyKovalik, I think I disagree with you about `0` to be an invalid value. The Date API defines that argument as "an integer value that is the number of milliseconds since 1 January 1970 UTC". So, there is nothing wrong with it.

Whether an application could not accept it in order to avoid some kind of problem, a programmer should document that with a test case. However, the (Jest-extended) library should allow it, since it is an acceptable value for a Date.

CoreyKovalik:
@thiagodp Was thinking along the lines of the same thing.  Would be nice to have a similar set of matchers for Unix timestamps

thiagodp:
@CoreyKovalik yes, it would be useful to have such matchers

[...]

PR #133, opened 2018-06-02
Add toBeValidDate matcher

tejasbubane:
Issue: #117

### What
Add new matcher `toBeValidDate`.

### Why
As suggested in the issue.

### Housekeeping

- [x] Unit tests


src/matchers/toBeValidDate/predicate.js
```diff
@@ -0,0 +1,7 @@
+let is = type => value => Object.prototype.toString.call(value) === `[object ${type}]`;
```
mattphillips: This might be a useful function to pull out into `utils` and then import here

src/matchers/toBeValidDate/predicate.js
```diff
@@ -0,0 +1,7 @@
...
+
+let hasDateType = is('Date');
+
+let isValidDate = value => hasDateType(value) && !isNaN(value) && !isNaN(value.getTime());
+
+export default expected => isValidDate(expected);
```
mattphillips: This can just be:

```js
export default isValidDate;
```

We don't need both the left and right hand sides of the invocation :)

src/matchers/toBeValidDate/predicate.js
```diff
@@ -0,0 +1,7 @@
+let is = type => value => Object.prototype.toString.call(value) === `[object ${type}]`;
+
+let hasDateType = is('Date');
```
mattphillips: Would it be possible to use `const` for consistency with other matchers?

mattphillips:
Love this @tejasbubane I've added a couple of minor nitpicks but looks awesome! Would you be able to merge master in too to fix the conflicts?

tejasbubane:
@mattphillips I have update both `toBeDate` and `toBeValidDate` matchers as per your comments.

Issue #139, opened 2018-06-25
Remove chalk to support browsers


frederickfogerty:
**Feature Request**

Description:

As jest is being used more and more in the browser, it would be great if this module could run in the browser. Currently this repo is blocking me due to its dependency on chalk. It seems that it's only used in [one place](https://github.com/jest-community/jest-extended/blob/3fdb02487c5bb5e4fa012c65c9eb519d963dc34e/src/index.js#L1)

Possible solution:

Remove the dependency on `chalk`.

mattphillips:
Closing as #140 is now merged :tada:

PR #140, opened 2018-06-25
feat: remove chalk to support browsers

frederickfogerty:
From #139

### What
Removes dependency on chalk. Right now the call to `red` has been removed. There is potential to run different code for this based on the platform, using a technique such as [this](https://github.com/chalk/chalk/issues/1#issuecomment-388428335), but I'm not sure how stable/supported this technique is.

### Why
Chalk does not work in the browser. As Jest is moving to support browsers, it is important that jest-extended can work in browser as well.

### Housekeeping

- [x] Unit tests


mattphillips:
Thanks @frederickfogerty given how little `chalk` was used this is a sensible change 👍

PR #143, opened 2018-06-30
Add toHaveBeenCalledAfter matcher for mocking functions.

benjaminkay93:
### What

I have added a toHaveBeenCalledAfter matcher.

### Why

The toHaveBeenCalledBefore matcher works great. However, the 
.not.toHaveBeenCalledBefore passes when called at the same time
which is true, but doesn't enable checking it has been called after.

The toHaveBeenCalledAfter matcher is a reverse of the logic in 
toHaveBeenCalledBefore, so will also pass on the same timestamp 
on .not.toHaveBeenCalledAfter but will enable to tests of mock1 being
called before mock2.

### Housekeeping

- [x] Unit tests


mattphillips:
Hey @benjaminkay93 this look great thanks a bunch dude 👍

Issue #150, opened 2018-07-30
"TypeError: Object.values is not a function" on Node 6


amercier:
**Bug**

- `package` version: 0.8.0
- `node` version: 6.14.3 (LTS, [end-of-life April 2019](https://github.com/nodejs/Release))
- `npm` (or `yarn`) version: 0.33.11

Relevant code or config

```javascript
require('jest-extended');
```

What you did:

Require `jest-extended` on Node 6.

What happened (please provide anything you think will help):

```
 FAIL  src/index.spec.js
  ● Test suite failed to run
    TypeError: Object.values is not a function
      at Object.<anonymous> (node_modules/jest-extended/dist/matchers/index.js:269:26)
      at Object.<anonymous> (node_modules/jest-extended/dist/index.js:3:17)
```

benjaminkay93:
Ahh, Object.values only got introduced in node 7, Looks like it got brought in with https://github.com/jest-community/jest-extended/pull/142

benjaminkay93:
We could swap this out for 
``` js
export default Object.keys(imports).map(key => imports[key]).reduce...
```

mattphillips:
Thanks @amercier for raising this, I've just published [v0.8.1](https://www.npmjs.com/package/jest-extended/v/0.8.1) with @benjaminkay93's fix :smile:

Issue #123, opened 2018-04-19
Add node 6 in travis


mattphillips:
Currently snapshots fail in travis when running in node 6. This is because the colours being used in pretty-format are striped away from the snapshot. There is a related issue in Jest (see https://github.com/facebook/jest/issues/5431), which once fixed should resolve this issue and revert 0d8ac8b.

dhuang612:
I would like to work on this

keeganwitt:
Node 6 is no longer a relevant target, so we can close this issue. We are also using GitHub Actions now instead of Travis.

PR #151, opened 2018-07-30
Move to Object.keys as opposed to Object.values

benjaminkay93:
### What
Bug

### Why
https://github.com/jest-community/jest-extended/issues/150 was raised today, Object.values doesn't work in Node 6, which is still in LTS from Node until next year. This uses Object.keys which is supported in node 6.

### Notes
I cannot test this in Node 6 as failure snapshot tests fail locally. However, the unit tests for the file charged passes.

### Housekeeping

- [x] Unit tests
- [ ]  Documentation is up to date

### DIFF
--- a/src/matchers/index.js
+++ b/src/matchers/index.js
@@ -2,4 +2,6 @@ import importAll from 'import-all.macro';
 
 const imports = importAll.sync('./*/index.js');
 
-export default Object.values(imports).reduce((acc, matcher) => ({ ...acc, ...matcher.default }), {});
+export default Object.keys(imports)
+  .map(key => imports[key])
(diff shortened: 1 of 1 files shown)


amercier:
In order to catch this earlier, would it be a good idea to add multiple NodeJS versions to `.travis.yml`? Like so:

```yml
node_js:
  - node # Latest
  - --lts # Current LTS
  - 6 # Maintenance LTS
```

The drawback is that it slows down builds a bit (even though they are running in parallel).

mattphillips:
Thanks @benjaminkay93 for fixing this so quickly!!

@amercier we were previously running node 6 in travis but because of #123 we had to remove it

Issue #158, opened 2018-08-14
toBeEmpty Map and Set


mattphillips:
Add support to `toBeEmpty` for `Map` and `Set` data structures

mattphillips:
Hey @moredip that sounds like a sensible approach! I like that it will also apply to custom iterables :smile:

I think we might be able to leverage the spread operator to condense the logic to something like:

```js
[...iterable].length === 0
```

Fancy sending a PR for this?

mattphillips:
Merged 🎉

PR #160, opened 2018-08-15
extend toBeEmpty to support Map, Set, and any other iterable

moredip:
### What

Teach toBeEmpty how to test for emptiness of any [Iterable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators#Iterables). 

### Why
Implements #158.  Adds support for testing emptiness of Map and Set, amongst others.

### Notes
@mattphillips [suggested](https://github.com/jest-community/jest-extended/issues/158#issuecomment-412944339) implementing this using `[...value].length === 0`. That would have been a neat, terse implementation but would not have worked for non-terminating iterables (which admittedly is a pretty obscure edge case). A more plausible advantage of not using the destructuring approach is that we avoid unecessary evaluation of potentially large/expensive iterables - we only need to go as far as yielding one value. Not really a huge deal either way, but it seemed a good tradeoff to make the implementation a bit bigger in exchange for supporting weird custom iterables.

### Housekeeping

- [x] Unit tests


mattphillips:
@moredip sorry for the delay in getting this merged! Looks great thanks 👍

PR #164, opened 2018-09-07
Add toBeIso8601Strict custom matcher (toBeIso8601 will come later down the line)

benjaminkay93:
### What - Feature
Added a strict matcher for the ISO8601 standard

### Why
When testing things, it could be helpful to check against the standard in the format that is generally the accepted defacto format -> 'YYYY-MM-DDThh:mm:ss.SSSZ'

### Notes
Can people read over this and check the logic in my regex test? They are fiddly things and I'm sure there are cases I've missed (Not handled leap years or for example 31st April)

### Housekeeping

- [x] Unit tests


OliverEvans96:
Any updates here?

imjordanxd:
Any progress on this? This would be a handy assertion to have.

rickhanlonii:
Hey, sorry for the delay here. This seems like a pretty nice use-case. Would recommend publishing this as a separate module.

blabute:
> Hey, sorry for the delay here. This seems like a pretty nice use-case. Would recommend publishing this as a separate module.

Hello! Are you suggesting that this should not be a part of `jest-extended`? I had thought that `jest-extended` was meant to include shared matchers like this?

jkillian:
I also just found myself searching through the jext-extended API docs for this functionality - would be great to have as objects being sent over the wire (or any JSON-serialized data really) are going to contain date strings instead of actual date objects.

maturanomx:
> Hey, sorry for the delay here. This seems like a pretty nice use-case. Would recommend publishing this as a separate module.

Why? Please reconsider include it

PR #163, opened 2018-09-07
Fix types for toSatisfy()/toSatisfyAll()

tony19:
### What
Bug fix in types

### Why
The type for the arg of `toSatisfy()`/`toSatisfyAll()` is currently a predicate `() => boolean`, but the predicate itself has an argument (the item being iterated). The current signature causes an error in this TypeScript code:

    expect(fooItem).toSatisfy((x: any) => x.myBoolProp); // Argument of type '(x: any) => any' is not assignable to parameter of type '() => boolean'.
    expect(fooArray).toSatisfyAll((x: any) => x.myBoolProp); // Argument of type '(x: any) => any' is not assignable to parameter of type '() => boolean'.

### Housekeeping

- [ ] Unit tests
- [ ] Documentation is up to date
- [ ] No additional lint warnings
- [x] Add yourself to contributors list (`yarn contributor`)
- [x] [Typescript definitions](https://github.com/jest-community/jest-extended/blob/master/types/index.d.ts) are added/updated where relevant

PR #165, opened 2018-09-09
Use generics for toSatisfy/toSatisfyAll

tony19:
### What
Feature (change in types)

### Why
In #163, we changed the `toSatisfy`/`toSatisfyAll` predicate to take an argument of type `any`, which can potentially hide type-related errors, such as:

    expect(promise1).toSatisfy(x => x.nonexistent()); // runtime error
    expect([promise1, promise2]).toSatisfyAll(x => x.nonexistent()); // runtime error

We can make the predicate generic to mitigate the problem, allowing for:

    expect(promise1).toSatisfy<CustomPromise>(x => x.isPending);
    expect([promise1, promise2]).toSatisfyAll<CustomPromise>(x => x.isPending);

The default type argument is `any` to maintain backward compatibility (i.e., it's still possible to call `toSatisfy(x => /*...*)` and `toSatisfyAll(x => /*...*)` without a type argument).

### Housekeeping

- [ ] Unit tests

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -39,7 +39,7 @@ declare namespace jest {
      * Use `.toSatisfy` when you want to use a custom matcher by supplying a predicate function that returns a `Boolean`.
      * @param {Function} predicate
      */
-    toSatisfy(predicate: (x: any) => boolean): R;
+    toSatisfy<T = any>(predicate: (x: T) => boolean): R;
 
     /**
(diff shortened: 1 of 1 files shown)


tony19:
It turns out specifying the type inline works just as well without any modification:

    .toSatisfy((x: CustomPromise) => x.isPending)

stilettk:
Is it possible to infer type from `expect` argument?

Issue #130, opened 2018-05-24
Add matchers for asserting that a promise has resolved or rejected (regardless of value)


Guardiannw:

benjaminkay93:
This has been sat here for a while, tbh I'm not sure how to go about this, nor how I feel about it either.
I say this because I'm on 10% time at work and figured I'm free enough to kick-start this conversation (hopefully).
https://github.com/mattphillips/babel-jest-assertions is potentially a solution, as this enforces the expect to be called in the advent of a none existent resolve/reject.

I was thinking can we tie into the timeout for a test in the jest config and expect it to be completed by the time that timeout comes, (i think 5 seconds default?)

SimenB:
@mattphillips ping 🙂

mattphillips:
Hey guys sorry for the delay in coming back to this, I’m currently holidaying and AFK for the most-part. 

@benjaminkay93 I think this should be possible with the new async matchers. I’d imagine it would be a similar implementation to Jest’s .resolves API.

benjaminkay93:
okay, im happy to give this a go if nobody else wants to, are there any examples of async custom matchers knockin about? would make life easier, ill ofc have a google too

benjaminkay93:
sweet cheers matt, enjoy your hols! 🏖

Morikko:
Would it be possible to return the error message when `toResolve()` fails ? 
Now, it is quite hard to debug without running extra manual tests.

kael89:
Hi there, congratulations for this awesome library! I was just browsing through this issue, and I have the impression it has since been implemented? Check [.toResolve()/.toReject()](https://github.com/jest-community/jest-extended#toresolve)

PR #168, opened 2018-09-21
Add toResolve/toReject custom matchers

benjaminkay93:
### What
Added custom matcher's for determining if a promise has resolved or rejected

### Why
As pair https://github.com/jest-community/jest-extended/issues/130

### Housekeeping

- [x] Unit tests

### DIFF
--- a/src/matchers/toReject/index.js
+++ b/src/matchers/toReject/index.js
@@ -3,17 +3,17 @@ import { matcherHint, printExpected, printReceived } from 'jest-matcher-utils';
 import predicate from './predicate';
 
 const passMessage = () => () =>
-  matcherHint('.not.toResolve') +
-  '\n\n' +
-  `Expected promise to ${printExpected('reject')}, however it ${printReceived('resolved')}.\n`;
-
-const failMessage = () => () =>
-  matcherHint('.toResolve') +
+  matcherHint('.not.toReject') +
   '\n\n' +
   `Expected promise to ${printExpected('resolve')}, however it ${printReceived('rejected')}.\n`;
 
+const failMessage = () => () =>
+  matcherHint('.toReject') +
+  '\n\n' +
+  `Expected promise to ${printExpected('reject')}, however it ${printReceived('resolved')}.\n`;
+
 export default {
-  toResolve: async promise => {
+  toReject: async promise => {
     const pass = await predicate(promise);
     if (pass) {
       return { pass: true, message: passMessage() };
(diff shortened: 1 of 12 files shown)


src/matchers/toReject/predicate.js
```diff
@@ -0,0 +1,10 @@
+export default async promise => {
```
mattphillips: I don't think the `async` is needed here as we aren't awaiting anything. Also the `resolved`/`rejected` values aren't being used, how about something like:

```js
export default promise => promise.then(() => false, () => true);
```

src/matchers/toReject/predicate.test.js
```diff
@@ -0,0 +1,13 @@
+import predicate from './predicate';
+
+describe('toResolve predicate', () => {
```
mattphillips: Should be `toReject predicate` 😄

src/matchers/toReject/index.js
```diff
@@ -0,0 +1,19 @@
+import { matcherHint } from 'jest-matcher-utils';
+
+import predicate from './predicate';
+
+const passMessage = () => () =>
```
mattphillips: No need for the extra function here or below for the `failMessage`

src/matchers/toResolve/index.js
```diff
@@ -0,0 +1,19 @@
+import { matcherHint } from 'jest-matcher-utils';
+
+import predicate from './predicate';
+
+const passMessage = () => () =>
```
mattphillips: Same here

src/matchers/toResolve/predicate.js
```diff
@@ -0,0 +1,10 @@
+export default async promise => {
```
mattphillips: Same as above comment :)

```js
export default promise => promise.then(() => true, () => false);
```

mattphillips:
@benjaminkay93 overall this looks awesome dude! Just a few minor comments above 😃

types/index.d.ts
```diff
@@ -283,6 +283,16 @@ declare namespace jest {
      */
     toBeSealed(): R;
 
+    /**
+     * Use `.toResolve` when checking if a promise resolves.
+     */
+    toResolve(): R;
```
mattphillips: Would you be able to add these types to the `expect` interface too (asymmetric matchers)

mattphillips:
Thanks @benjaminkay93 :)

types/index.d.ts
benjaminkay93: aha I will have a nosey at the expect now, thanks matt 🎉

Issue #157, opened 2018-08-13
toThrow(constructor, message)


mattphillips:
Clone of https://github.com/facebook/jest/issues/3659

```js
expect(()=>{
  throw new Error('1234');
}).toThrow(Error, 'hello');
```

All that needs working out is the name as `toThrow` is built into core and so this shouldn't clash with it

benjaminkay93:
I like the idea of toThrowWithMessage

mattphillips:
@natealcedo yup sounds good to me 👍

natealcedo:
So I've actually started taking a look at how to implement this. I thought it would be a good idea to copy what jest's default `toThrow` matcher currently does.  On that note I have a couple of questions. 

1. Should we be doing something like this `toThrowWithError(error, message)` where `message` can either be a regex or a string as per what jest's `toThrow` does?
2. How strict do we want to be with argument validation. Should we raise an error when `message` has not been provided?

Just some of the thoughts I've been considering. I'll try to get in an initial pull request sometime next week so that you guys can verify what I've worked on and whether I'm going in the right direction.

mattphillips:
@natealcedo I like the idea keeping this  close to Jest’s current API with regex support. 

As for the validation we should probably fail the matcher  with a nice error message, saying the argument is missing. 

At some point in the future I’m planning on adding validation throughout all of the matchers so don’t overthink it for now, perhaps just checking the message is supplied :smile:

natealcedo:
Let’s close since #170 fixes this :)

mattphillips:
Released as part of [v0.11.0](https://github.com/jest-community/jest-extended/releases/tag/v0.11.0) :tada:

PR #170, opened 2018-09-22
Add a toThrowWithMessage matcher

natealcedo:
### What
This pull request adds in a new matcher for asserting that a callback function throws an error of a given type and given error message.

### Why
This feature has been requested in core but after some discussion, it was decided that implementing it here would be the better solution. 

This PR fixes the following issues
1. https://github.com/facebook/jest/issues/7000
2. https://github.com/facebook/jest/issues/3659
3. #157 

### Notes
I've gone down the road of testing the matcher both by extending expect with this matcher and by invoking the matcher itself. Since this repository doesn't have some of the util tools that the Jest repo has, namely [runJest](https://github.com/facebook/jest/blob/master/e2e/runJest.js), I am not able to test the failing scenarios since I can't capture whatever is logged to stdErr to assert that the error message is indeed correct. As such, I've just invoked the matcher itself and asserted whether it passes or not and used snapshot testing for asserting error messages.

Also, the goal for me in this matcher was to stick as close to the default `toThrowError`matcher. As such, this matcher also accepts both a `string` and `regex` for matching the error message. 

The code is abit verbose right now but I'm more than happy to have a discussion on how to amend it moving forward.
### Housekeeping

- [x] Unit tests


README.md
```diff
@@ -398,6 +399,30 @@ test('passes when value is a function', () => {
 });
 ```
 
+#### .toThrowWithMessage(type, message)
+
+Use `.toThrowWithMessage` when checking if a callback function throws an error with a given error type and given error message.
```
mattphillips: Perhaps it's worth mentioning that the message can be either a `String` or `RegExp`?

types/index.d.ts
```diff
@@ -337,5 +337,13 @@ declare namespace jest {
...
      */
     toIncludeMultiple(substring: string[]): R;
+
+    /**
+     * Use `.toThrowWithMessage` when checking if a callback function throws an error of a given type with a given error message.
+     *
+     * @param {Function} type
+     * @param {String | RegExp} message
+     */
+    toThrowWithMessage(type: Function, message: string | RegExp): R;
```
mattphillips: This needs adding to the `expect` interface too

src/matchers/toThrowWithMessage/index.js
```diff
@@ -0,0 +1,98 @@
...
+Thrown:
+${printReceived(received)}
+`;
+
+export default {
+  toThrowWithMessage: (callback, type, message) => {
+    if (!callback || typeof callback !== 'function') {
+      return {
+        pass: false,
+        message: () => `
```
mattphillips: The spacing of these validation errors is tabbed in too much for the printouts (see the snapshots)

mattphillips:
@natealcedo this is looking good just a few small comments :)

src/matchers/toThrowWithMessage/index.js
natealcedo: Alright, I've pushed in a change for this. Do take a look.

types/index.d.ts
mattphillips: Ah we just merged the types for asymmetric matchers so needed to update this file twice - all good though I just did it for you :)

mattphillips:
Thanks @natealcedo 👍

PR #184, opened 2018-11-23
Fix typings for toContainEntry, toContainEntries, toContainAllEntries and toContainAnyEntries

tleunen:
Having the entries to be `[string, string]` is a bit too limited.
Most the the time, the keys will be strings. Although they could be Symbols too now. But especially for the values where it could be anything...

### Why
Typescript complaints when we try to use one of these functions with the entries of an object where the keys or values are not strings.

### Housekeeping

- [ ] Unit tests

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -252,28 +252,28 @@ declare namespace jest {
      *
      * @param {Array.<String, String>} entry
      */
-    toContainEntry(entry: [string, string]): R;
+    toContainEntry<T>(entry: [keyof T, T[keyof T]]): R;
 
     /**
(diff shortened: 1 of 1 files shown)


tleunen:
@mattphillips - I know you're busy. But I'll appreciate if you can take a look when you have some time. No rush though.

mattphillips:
@tleunen dude my  bad leaving this so long to merge!! This is a really nice change, that I can now appreciate as I'm using TS thanks a bunch. I've just published this under [v0.11.1](https://www.npmjs.com/package/jest-extended/v/0.11.1)

PR #190, opened 2019-01-16
Missing Typescript import info in README.md

distante:
doc: Missing typescript information in README.md

To `jest-extended` to work with Typescript files an import is needed but this info was missing in the README.md


mattphillips:
Thanks for the PR @distante, however @SimenB is correct this has already been covered :smile:

PR #202, opened 2019-03-03
feat: Add .toLog() matcher

ollelauribostrom:
This might need a bit of tweaking so consider it a WIP. Thought I put it up here to see if something like this would be useful, input is much appreciated 🙂 

### What
This PR adds a new matcher for asserting that a callback function outputs a message to the console. The given console method is mocked when running the callback, so nothing is actually printed to the console. 

### Why
I sometimes find myself wanting to assert that some function prints a message to the console. Like for instance checking that some information is printed, or that a deprecation warning is printed when calling a certain method. This matcher would be convenient for such scenarios.

Instead of doing something like:
```js
const log = jest.spyOn(console, 'log');
log.mockImplementation(() => {});
someFunction();
expect(log).toHaveBeenCalledWith('some message');
log.mockRestore();
```
We could do
```js
expect(someFunction).toLog('some message');
expect(someOtherFunction).toLog('some deprecation warning', 'warn');
```
### Housekeeping
- [x] Unit tests

### DIFF
--- a/src/matchers/toLog/predicate.js
+++ b/src/matchers/toLog/predicate.js
@@ -1,0 +1,21 @@
+import { equals } from 'expect/build/jasmine_utils';
+
+export default (fn, expected, consoleMethod) => {
+  const spy = jest.spyOn(console, consoleMethod);
+  spy.mockImplementation(() => {});
+  fn();
+  spy.mockRestore();
+  if (spy.mock.calls.length === 0) {
+    return { pass: false };
+  }
+  const actual = spy.mock.calls;
+  let pass = false;
+  actual.forEach(call => {
+    call.forEach(arg => {
+      if (equals(arg, expected)) {
+        pass = true;
+      }
+    });
+  });
+  return { actual, pass };
+};
--- a/README.md
+++ b/README.md
@@ -56,6 +56,8 @@ If you've come here to help contribute - Thanks! Take a look at the [contributin
     - [.toBeBoolean()](#tobeboolean)
     - [.toBeTrue()](#tobetrue)
     - [.toBeFalse()](#tobefalse)
+  - [Console](#console)
+    - [.toLog()](#tolog)
   - [Date](#date)
     - [.toBeDate()](#tobedate)
     - [.toBeValidDate()](#tobevaliddate)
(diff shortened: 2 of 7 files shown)


src/matchers/toLog/predicate.js
```diff
@@ -0,0 +1,21 @@
+import { equals } from 'expect/build/jasmine_utils';
+
+export default (fn, expected, consoleMethod) => {
+  const spy = jest.spyOn(console, consoleMethod);
```
benjaminkay93: if we set up the spy here I can imagine we could incure larger performance implications than if it wasn't set at run time, but ill comment on that in my main response 😇

ollelauribostrom:
Thanks a lot for the input @benjaminkay93, really like having an explicit matcher for each console method (PR is updated to reflect this). A few thoughts,

- Should I put each matcher in its own directory?
- Not sure I get what you mean by setting up the spies on initiation rather than run time. Wouldn't that mean that we would always be spying on the console methods?

williamrchr:
Nit: I would adjust the matcher method to be more human-readable:
`.toLog()`
`.toLogError()`
`.toLogInfo()`

williamrchr:
> Thanks a lot for the input @benjaminkay93, really like having an explicit matcher for each console method (PR is updated to reflect this). A few thoughts,
> 
>     * Should I put each matcher in its own directory?
> 
>     * Not sure I get what you mean by setting up the spies on initiation rather than run time. Wouldn't that mean that we would always be spying on the console methods?

> Should I put each matcher in its own directory?
Will let the project maintainer determine, but current structure would indicate that each should have its own directory.

> Spying on initiation rather than run time
I don't think you need a spy here. You could simply wrap `console.log()` and friends and have a counter that adds everytime something gets added. Then there's not an internal dependency on jest.  See: https://stackoverflow.com/questions/19846078/how-to-read-from-chromes-console-in-javascript

ollelauribostrom:
@williamrchr Thanks for having a look, did some more updates to this  🙂

AmyFoxFN:
Hi, all
Thanks for the design and endeavor for the console matchers. It is really helpful to me and my team. In fact , we are just looking for something like this.

But there is a question, Why this process stoped and Maybe when will this PR is probably be merged?

Look forward to knowing more about this, and Thanks a lot!

keeganwitt:
`README.md` and `types/index.d.ts` both need updated to reflect the separate functions that were added.

src/matchers/toLog/predicate.js
```diff
@@ -0,0 +1,17 @@
+import { equals } from 'expect/build/jasmine_utils';
```
keeganwitt, line 1: I assume this path used to be correct, but it's now `jasmineUtils` instead of `jasmine_utils`.

PR #203, opened 2019-03-04
Validate if spy or mock was passed when expected

atsikov:
### What

Add error to `toHaveBeenCalledBefore` and `toHaveBeenCalledAfter` matchers in case passed params are not instances of `jest.Mock`

### Why

Jest has similar errors when non spy or mock is passed to `toHaveBeenCalled` or other mock-related matchers.

- [x] Unit tests


src/utils/index.js
```diff
@@ -8,4 +8,8 @@ export const determinePropertyMessage = (actual, property, message = 'Not Access
   return actual && actual[property] ? actual[property] : message;
 };
 
+export const isJestMockOrSpy = obj => {
+  return !!obj._isMockFunction && typeof obj.mock !== 'undefined';
```
benjaminkay93: I'm having trouble reading this logic 🙊, could you break it out to be more explicit so simple minds like mine can parse this easier please 😄

src/matchers/toHaveBeenCalledBefore/index.js
```diff
@@ -18,8 +20,24 @@ const failMessage = (firstInvocationCallOrder, secondInvocationCallOrder) => ()
   'Received second mock with invocationCallOrder:\n' +
   `  ${printReceived(secondInvocationCallOrder)}`;
 
+const mockCheckFailMessage = (value, isFirstValue) => () =>
+  matcherHint('.toHaveBeenCalledAfter') +
+  '\n\n' +
+  `Matcher error: ${isFirstValue ? 'first' : 'second'} value must be a mock or spy function\n` +
```
benjaminkay93: im not sure on `first` and `second` being the right wording here, could we potentially cause ambiguation for users if they read `second value must be a mock or spy function` and presume this means they have to give a second value to the matcher, not that the matchers first value isn't a mock/spy?

src/matchers/toHaveBeenCalledAfter/index.js
```diff
@@ -18,8 +20,24 @@ const failMessage = (firstInvocationCallOrder, secondInvocationCallOrder) => ()
   'Received second mock with invocationCallOrder:\n' +
   `  ${printReceived(secondInvocationCallOrder)}`;
 
+const mockCheckFailMessage = (value, isFirstValue) => () =>
+  matcherHint('.toHaveBeenCalledAfter') +
+  '\n\n' +
+  `Matcher error: ${isFirstValue ? 'first' : 'second'} value must be a mock or spy function\n` +
```
benjaminkay93: See comment on `toHaveBeenCalledBefore` Matcher 😃

src/matchers/toHaveBeenCalledBefore/index.js
```diff
@@ -1,5 +1,7 @@
 import { matcherHint, printExpected, printReceived } from 'jest-matcher-utils';
 
+import { isJestMockOrSpy } from '../../utils';
```
benjaminkay93: I really like how you broke this out to the utils 💯

src/matchers/toHaveBeenCalledBefore/index.js
atsikov: Yes, I also spent some time figuring out how to word it better. Probably `actual` and `expected` instead of `first` and `second`?

src/utils/index.js
atsikov: Well, not sure how to break it. `_isMockFunction` is a boolean prop which jest mocks and spies have. Also since matchers use values from `mock` object, I added check for this. Probably `typeof obj.mock === "object"` would be more explicit.

And in case this  `!!` confuses you, it can be safely removed 😄  Just since I mostly work with typescript, I usually add this explicit boolean cast.

src/matchers/toHaveBeenCalledBefore/index.js
atsikov: Thanks! I usually try to keep such atomic functions reusable.

src/utils/index.js
atsikov: Please check if it became better now

src/matchers/toHaveBeenCalledBefore/index.js
atsikov: Replaced with "Received" and "Expected" with red and green color, just as in matcher hint.

src/matchers/toHaveBeenCalledAfter/index.js
atsikov: Updated as in `toHaveBeenCalledBefore`

src/utils/index.js
atsikov: Ugh, formatted it with line-breaks, but `prettier` knows better 😞

PR #204, opened 2019-03-04
Add Pass & Fail matchers

williamrchr:
This PR creates the implementation for the pass and fail matchers.

These matchers are currently unimplemented.

I'm not sold on the default pass/failure message, and I created separate pass/fail message constants for both matchers in case someone wants to update one but not the other. Let me know if you'd rather avoid the code duplication.

### Housekeeping

- [y] Unit tests
- [y ] Documentation is up to date
- [y] No additional lint warnings
- [y] Add yourself to contributors list (`yarn contributor`)
- [y] [Typescript definitions](https://github.com/jest-community/jest-extended/blob/master/types/index.d.ts) are added/updated where relevant

### DIFF
--- a/src/matchers/fail/index.js
+++ b/src/matchers/fail/index.js
@@ -1,0 +1,21 @@
+import predicate from './predicate';
+
+const passMessage = message => {
+  if (message) return () => message;
+  else return () => 'Failure!';
+};
+
+const failMessage = message => {
+  if (message) return () => message;
+  else return () => 'Failure!';
+};
+
+export default {
+  fail: (expected, message) => {
+    const pass = predicate(message);
+    if (pass) {
+      return { pass: true, message: passMessage(message) };
+    }
+    return { pass: false, message: failMessage(message) };
+  }
+};
--- a/src/matchers/pass/index.js
+++ b/src/matchers/pass/index.js
@@ -1,0 +1,21 @@
+import predicate from './predicate';
+
+const passMessage = message => {
+  if (message) return () => message;
+  else return () => 'Passed!';
+};
+
+const failMessage = message => {
+  if (message) return () => message;
+  else return () => 'Passed!';
+};
+
+export default {
+  pass: (expected, message) => {
+    const pass = predicate(message);
+    if (pass) {
+      return { pass: true, message: passMessage(message) };
+    }
+    return { pass: false, message: failMessage(message) };
+  }
+};
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -450,7 +450,16 @@
       "contributions": [
         "code"
       ]
-    }
+    },
+        {
+          "login": "williamrchr",
+          "name": "William Archer",
+          "avatar_url": "https://avatars2.githubusercontent.com/u/7919424?s=460&v=4",
+          "profile": "https://github.com/williamrchr",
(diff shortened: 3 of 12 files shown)


src/matchers/fail/index.js
```diff
@@ -0,0 +1,10 @@
+import predicate from './predicate';
+
+const failMessage = message => {
+  if (message) return () => message;
+  else return () => 'Failure!';
```
benjaminkay93: Should we have this be something a little bit more descriptive, e.g. 'failed by .fail() assertion.'

src/matchers/pass/index.js
```diff
@@ -0,0 +1,10 @@
+import predicate from './predicate';
+
+const passMessage = message => {
+  if (message) return () => message;
+  else return () => 'Passed!';
```
benjaminkay93: Should we have this be something a little bit more descriptive, e.g. 'passed by .pass() assertion.'

src/matchers/fail/index.js
benjaminkay93: also it should really be something that is immerative.
Following a similar standard to how git commits should be, (how i like to write test names anyways 😄 
[See this doc](https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53#a-properly-formed-git-commit-subject-line-should-always-be-able-to-complete-the-following-sentence)
for example: 'The test ...'

so in this situation you would read 'The test failed by .fail() assertion.'

williamrchr: Do you mean imperative mood? Wouldn't it be `fails by .fail() assertion`? I don't think there's a past tense imperative in English, but I'm not a linguist. happy to change it to whatever you'd prefer!

src/matchers/pass/index.js
williamrchr: Happy to change it to whatever you'd like!

src/matchers/fail/index.js
benjaminkay93: i think your right? `fails` make more sense than `failed`, same with `passed` and `passes` too do you think?

williamrchr: okay, so I'll change default message to `fails by .fail() assertion` and `passes by .pass() assertion`?

benjaminkay93: yeah, that's perfect 😄

williamrchr: updated!

DimaDK24:
Hi, guys! Why these matchers are still not merged? They are really great and useful matchers 😀

williamrchr:
Wish I could! The maintainer of this project is MIA and contributors can't do their thing.

DimaDK24:
@mattphillips, I hope you're not missing :)

rulrok:
Guys, I'm missing this feature.
@mattphillips  Let's integrate this D:

WORMSS:
I would suggest setting the return type of `.fail()` to `never` eg
```ts
/**
   * Failing assertion
   *
   * @param {String} message
   */
  fail(message: string): never;
```
This will let typescript know everything after is unreachable code.
eg
```ts
if ( something === null ) {
  expect().fail('something is null');
}
someFunctionThatNeedsAValue(something); // << know something is not null
```

PR #193, opened 2019-02-04
Update docs to use setupFilesAfterEnv

abdelrahman-elkady:
### What

**Feature** update docs to use `setupFilesAfterEnv`, resolves #192 

### Why
`setupTestFrameworkScriptFile` is deprecated in jest `24.0.0`, see #192

### Housekeeping

- [x] Unit tests
- [x] Documentation is up to date
- [x] No additional lint warnings
- [x] Add yourself to contributors list (`yarn contributor`)
- [x] [Typescript definitions](https://github.com/jest-community/jest-extended/blob/master/types/index.d.ts) are added/updated where relevant

PR #207, opened 2019-03-12
Update instructions to use setupFilesAfterEnv

joaomilho:
### What

Current README instructs users to use Jest's `setupTestFrameworkScriptFile` config, which is not updated. The `setupFilesAfterEnv` config allows for multiple files, so the rest of the explanation is unnecessary.

### Why

Cause the instructions it will not work in latest jest.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,25 +127,7 @@ Add jest-extended to your Jest setupTestFrameworkScriptFile configuration. [See 
 
 ``` json
 "jest": {
-  "setupTestFrameworkScriptFile": "jest-extended"
-}
-```
-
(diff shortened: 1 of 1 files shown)


mkgn:
I have added jest-extended to setupFilesAfterEnv as below;

setupFilesAfterEnv: ["./__tests__/setup.js","jest-extended"]

since I use test setups as well. Then I did a simple test:

expect(response.headers['content-type']).toStartWith('application/json');

But when run it says:

Property 'toStartWith' does not exist on type 'Matchers<any>'.

I tried switching the place of jest-extended in setupFilesAfterEnv array. still the same.

mattphillips:
Thanks for the PR @joaomilho I'm going to close this as it has been fixed as part of #193 / #215

PR #209, opened 2019-03-31
Add toRejectWith

EyalPerry:
### What
Added the `toRejectWith` promise matcher.

### Why
Consider the below function:

```javascript
   const underTest = async (paramA, paramB) => {
      const stepOne = await mockedDependency(paramA);
      const stepTwo = await unmockedDependency(stepOne, paramB);

      if(someCondition(stepTwo)){
           throw new Error('some invariant was not met');
      }

      return { key: stepOne.someKey, value: stepTwo } 
  };
```

#### Scenario
I would like to ensure that the `underTest` function throws when `someCondition` is met.
For this purpose, `jest-extended` introduces the `toReject` matcher, which introduces, IMHO- a testing anti pattern.

```javascript
await expect(underTest('a','b')).toReject();
```

#### Problem
`toReject` will pass on any on the following conditions:

- due to test misconfiguration, `mockedDependency` returns undefined:
causing `unmockedDependency` to throw.

- due to test misconfiguration, `mockedDependency` returns undefined:
`unmockedDependency` handles the undefined value properly, but computing the return value will throw when dereferencing `stepOne.someKey`.

- the test is properly configured, but `unmockedDependency` has a bug which causes it to throw.

This, in turn- might cause an unwary developer to believe that all is well and that the invariants are correctly guarded- when this is not necessarily the case.

#### Solution

The problem with `toReject` is that it does not make developers explicitly specify the expected cause for failure.

This PR introduces an alternative for `toReject` - `toRejectWith`:.

```javascript
const messageComparer = (a, b) => a.message === b.message;
const expectedError = new Error('some invariant was not met');

await expect(underTest('a','b')).toRejecWith(expectedError, messageComparer);
```

#### Design Decisions
- Uses `Object.is` under the hood when a comparer function is not specified because:
  - Anything in JS is throwable.
  - Testing mocked functions which throw a dummy error is supported out of the box.

- For more complex, user specific scenarios:
A comparer function can be passed as a second parameter.

### Notes

- I honestly believe `toReject` should be depracated due to the aforementioned reasons.

- This matcher has no`predicate.js` module, since it enables passing in a custom 'error comparer' parameter which defaults to Object.is.
Hope this is acceptable, let me know if it isn't.

- `yarn contributor` failed to work on my mac, so I added myself manually.
If this isn't ok, let me know how I can address this issue.

- Pretty sure I am incorrectly handling the edge case when `.not` is used with this matcher and no error is specified. 
Not sure regarding the correct way to address this.
Currently, there's no test for this.
I think it doesn't even make sense to use `.not` with this matcher.
Is there a way to prevent people from using it in that context?

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -450,6 +450,17 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login":"EyalPerry",
+      "name": "Eyal Perry",
+      "avatar_url": "https://avatars1.githubusercontent.com/u/9663481",
+      "profile": "https://stackoverflow.com/users/story/2223556",
+      "contributions": [
+        "code",
+        "doc",
+        "test"
+      ]
     }
   ],
   "repoType": "github"
--- a/README.md
+++ b/README.md
@@ -97,6 +97,7 @@ If you've come here to help contribute - Thanks! Take a look at the [contributin
   - [Promise](#promise)
     - [.toResolve()](#toresolve)
     - [.toReject()](#toreject)
+    - [.toRejectWith(error,comparer)](#torejectwitherror--comparer)
   - [String](#string)
     - [.toBeString()](#tobestring)
     - [.toBeHexadecimal(string)](#tobehexadecimal)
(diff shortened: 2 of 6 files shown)


mattholmes-moovel:
Any particular reason this has stalled? Anything I can do to help? The author of this PR is spot on, 'toReject' can cause way too many false positives without having something to confirm the rejection reason. This is something we'd really like to use in our code.

pdxmholmes:
Any particular reason this has stalled? Anything I can do to help? The original PR author is spot on, toReject() is far too loose and easy to write false positives. Having an extended method to be able to check the reject reason allows people to write much more directed tests.

If I can help to get this moving somehow please let me know.

imjordanxd:
Any progress on this?

PR #252, opened 2019-11-27
Add toBeDateString matcher

idan-at:
Added a new matcher to check whether a string is in a date format.

A use-case example: when serializing a JS object with date values, those become strings, and when comparing against the returned json, it can be useful to use this matcher.

### Housekeeping

- [x] Unit tests


bompus:
This is likely a dupe of #164 and would be super useful.

SimenB (approved):
thanks!

PR #257, opened 2019-12-11
Fix determinePropertyMessage() to return property if property is falsy

jblandry:
### What
Fix `determinePropertyMessage()` to return property if property is falsy, instead of "Not Accessible".

### Why
When using `.toBeArrayOfSize()` on an empty array it fails saying the property length is "Not Accessible" when it should say it is not the right length.

What it does:
```js
expect([]).toBeArrayOfSize(1);

//    expect(received).toBeArrayOfSize(expected)
//
//    Expected value to be an array of size:
//      1
//    Received:
//      value: []
//      length: "Not Accessible"
```

What it should do:
```js
expect([]).toBeArrayOfSize(1);

//    expect(received).toBeArrayOfSize(expected)
//
//    Expected value to be an array of size:
//      1
//    Received:
//      value: []
//      length: 0
```

### Notes
Corrected the test about custom message since functions have a `length` property, and added a new test for the specific case mentioned.

### Housekeeping

- [X] Unit tests
- [X] [Typescript definitions](https://github.com/jest-community/jest-extended/blob/master/types/index.d.ts) are added/updated where relevant

### DIFF
--- a/src/matchers/toBeArrayOfSize/__snapshots__/index.test.js.snap
+++ b/src/matchers/toBeArrayOfSize/__snapshots__/index.test.js.snap
@@ -7,7 +7,7 @@ Expected value to not be an array of size:
   <green>0</>
 Received:
   value: <red>[]</>
-  length: <red>\\"Not Accessible\\"</>"
+  length: <red>0</>"
 `;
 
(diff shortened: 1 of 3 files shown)


SimenB:
I'm not allowed to push to this fork, so I opened up #360

PR #283, opened 2020-05-06
Update docs section about typescript setup

### LINKED ISSUE #282: Update typescript setup docs after breaking change in ts-jest 25.5.0
In ts-jest `25.5.0` there was a [breaking change](https://github.com/kulshekhar/ts-jest/blob/master/CHANGELOG.md#breaking-changes) that now requires to list any additional typings in the `files` property of the `tsconfig.json`. The docs should be updated accordingly.

warrify:
### What
Updated the docs section to include the setup for `ts-jest >= 25.5.0`

### Why
`ts-jest >= 25.5.0` introduces [breaking changes](https://github.com/kulshekhar/ts-jest/blob/master/CHANGELOG.md#breaking-changes) that require additional configuration.

### Notes
Closes #282 

### Housekeeping

- [x] Unit tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,22 @@ If your editor does not recognise the custom `jest-extended` matchers, add a `gl
 import 'jest-extended';
 ```
 
+_Note: When using `ts-jest >= 25.5.0`_
+
+Since the [breaking changes]() in `25.5.0` you also need to update you `tsconfig.json` to include the new `global.d.ts` file in the `files` property like so:
+
+```json
+{
+  "compilerOptions": {
+    ...
+  },
+  ...
+  "files": ["global.d.ts"]
+}
+```
+
+Also note that when adding this for the first time this affects which files are compiled by the typescript compiler and you might need to add the `include` property as well. See the [typescript docs](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html) for more details.
+
 ## Asymmetric matchers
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -167,6 +167,22 @@ If your editor does not recognise the custom `jest-extended` matchers, add a `gl
 import 'jest-extended';
 ```
 
+_Note: When using `ts-jest >= 25.5.0`_
+
+Since the [breaking changes]() in `25.5.0` you also need to update you `tsconfig.json` to include the new `global.d.ts` file in the `files` property like so:
```
Livven, line 172:
```suggestion
Since the [breaking changes]() in `25.5.0` you may also need to update your `tsconfig.json` to include the new `global.d.ts` file in the `files` property like so:
```
Works just fine for me with `jest-extended@0.11.5` and `ts-jest@26.1.4` so I would suggest adding a "may" here.

In fact looking at https://github.com/kulshekhar/ts-jest/pull/1562 it seems that the breaking change only applies if you were using the `include` section in tsconfig.json previously.

villasv:
I also think it would be nice to warn users about declaration merging not working if you import the module.

In my case, I was using `import {} from '@jest/globals'` and this would remove the effect of the imported types from `jest-extended`. It only worked for me if I relied on the globally injected variables from jest.

SimenB:
> I also think it would be nice to warn users about declaration merging not working if you manually import the jest module as the jest docs comment: `import { expect } from '@jest/globals`.
> 
> It fails because `jest-extended` will merge its type declarations into the types provided by `@types/jest`. So if you use ` expect` from `@jest/globals`, the added matchers won't be there.

Agreed, wanna open up a PR @villasv? 🙂

PR #298, opened 2020-10-30
Allow any iterable in toInclude*Members

bogdanb:
### What

Modifies the three `toInclude{All,Any,Some}Members` matchers to also match on iterables.

### Why

This is meant mostly for testing things that return `Set` and `Map` instances, but it’s generally useful for any Iterable.

### Notes

I’m not sure exactly what’s wrong with the typing file, I get `TS2304: Cannot find name 'Iterable'.` errors.

### Housekeeping

- [x] Unit tests

### DIFF
--- a/src/matchers/toIncludeAnyMembers/index.test.js
+++ b/src/matchers/toIncludeAnyMembers/index.test.js
@@ -60,7 +60,7 @@ describe('.not.toIncludeAnyMembers', () => {
   test('fails when given array contains object value', () => {
     expect(() => expect([{ foo: 'bar' }]).not.toIncludeAnyMembers([{ foo: 'bar' }])).toThrowErrorMatchingSnapshot();
     expect(() =>
-      expect(new Set([{ foo: 'bar' }])).not.toIncludeAnyMembers([{ foo: 'bar' }])
+      expect(new Set([{ foo: 'bar' }])).not.toIncludeAnyMembers([{ foo: 'bar' }]),
     ).toThrowErrorMatchingSnapshot();
   });
 


bogdanb:
I’m not actually sure where to update the documentation. I updated the comments in the TypeScript definition file (to say Iterable instead of Array), but I’m not sure if there’s anything else to update.

The type definitions show errors, though, on the type of the argument (which can also be an Iterable, now), so maybe this also needs some config update to see the Iterable definition?

SimenB:
>I’m not actually sure where to update the documentation. I updated the comments in the TypeScript definition file (to say Iterable instead of Array), but I’m not sure if there’s anything else to update.

Could you update the docs? It currently says "`Array`", and maybe add an example using a `Set`?

keeganwitt:
This won't work for maps. `contains()` uses `findIndex()`, which won't work for maps.

One way to test whether something is iterable is with
```javascript
function isIterable(obj) {
  // checks for null and undefined
  if (obj == null) {
    return false;
  }
  return typeof obj[Symbol.iterator] === 'function';
}
```
But this also returns true for a `string` -- do we think that's expected behavior for these functions?

WORMSS:
> But this also returns true for a `string` -- do we think that's expected behavior for these functions?

Since string IS an iterable, it would be weird if it said false.. 
Is there advantage of a `if null` check instead of doing 
```ts
return typeof obj?.[Symbol.iterator] === 'function';
```

keeganwitt:
> Since string IS an iterable, it would be weird if it said false..

I agree, I just wasn't sure that was something considered when this was opened.

> Is there advantage of a if null check instead of doing
> ```
> return typeof obj?.[Symbol.iterator] === 'function';
> ```

Nope, just didn't think of it when I grabbed that example.

PR #313, opened 2021-04-19
Move @types/jest to be a dependency

idan-at:
Bug fix: TS can't compile unless `@types/jest` is somehow hoisted by another package.

Since jest-extended uses the jest types on its d.ts file, which is packed on the tarball, it actually requires @types/jest as a production dependency, and not a dev-dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,6 @@
   "license": "MIT",
   "repository": "jest-community/jest-extended",
   "devDependencies": {
-    "@types/jest": "^23.3.2",
     "babel-cli": "^6.26.0",
     "babel-core": "^6.26.0",
     "babel-eslint": "^8.0.1",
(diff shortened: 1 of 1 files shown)


SimenB:
It should be an optional peer dependency probably...

idan-at:
It can be a peer dependency, but I don't think it can be optional because TS won't compile.

From what I know, in TS when you expose a d.ts that relies on external types, those should be production dependencies, since those types should be present when your package is being installed.

I ended up fixing the issue by installing the dependency myself (which suggests a peer dependency will work), but IMO making everyone that uses this package and TS to install the jest types isn't the most convenient

SimenB:
Right, but then yarn/npm would tell you to install it, which makes more sense than this module bringing its own. Why would you use Jest, TypeScript and this module, but not install `@types/jest` yourself?

idan-at:
My use case was a shared library of matchers, which re-exports jest-extended and other matcher libraries (like `jest-json`).

For every artifact, we install `jest` (with types) and this library, but this library alone doesn't have to use jest, if it doesn't have tests 😅

PR #326, opened 2021-06-23
Replace Travis with GitHub Actions

keeganwitt:
### What
Switching from Travis to GitHub Actions.

### Why
Travis [changed](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) their pricing model, which allocates 10,000 credits/minutes for open source projects. Once those credits are consumed, you must repeat the process to request more credits.

### Notes
* GitHub Free [offers](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions) 2,000 credits per month, but there is no total cap and no renewal process is needed. If we have spikey months, we may not wish to proceed with this change.
* Codecov authentication isn't configured yet because I didn't have that information.


SimenB:
GH actions are free for open source (public repos), no limits

image

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,22 @@
...
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+jobs:
+  ci:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [12.x, 14.x, 16.x]
```
SimenB, line 12: we currently support node 8. I'm down with just supporting the current LTSes, tho 👍

package.json
```diff
@@ -35,11 +35,11 @@
   "license": "MIT",
   "repository": "jest-community/jest-extended",
   "devDependencies": {
-    "@types/jest": "^23.3.2",
+    "@types/jest": "^24.0.0",
     "babel-cli": "^6.26.0",
     "babel-core": "^6.26.0",
     "babel-eslint": "^8.0.1",
-    "babel-jest": "^23.0.1",
+    "babel-jest": "^24.0.0",
```
SimenB, line 42: ?

package.json
```diff
@@ -99,7 +99,7 @@
       "babel-plugin-transform-es2015-modules-commonjs",
       "transform-object-rest-spread",
       "transform-async-to-generator",
-      "babel-jest-assertions",
+      "./node_modules/babel-jest-assertions",
```
SimenB, line 102:
```suggestion
      "module:babel-jest-assertions",
```

?

SimenB (approved):

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,22 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - run: yarn install
+      - run: yarn lint
+      - run: yarn test:coverage
```
SimenB, line 21: should be enough to add https://github.com/codecov/codecov-action

keeganwitt:
> GH actions are free for open source (public repos), no limits

Ah, I thought public repos fell under the "free" tier of 2,000 minutes per month. But now I see what you're saying

> GitHub Actions usage is free for both public repositories and self-hosted runners.

Even better!

package.json
keeganwitt: Because I bumped Jest to 24 so I could use `--color`, I also bumped this dependency. Both are dev dependencies.

.github/workflows/ci.yaml
keeganwitt: Ah. I didn't realize that. I like the idea of supporting only current LTS, but what does Jest itself officially support? Or is that what you meant by "we"?

package.json
keeganwitt: I thought that was a Babel 7 syntax? If I'm mistaken, then yea, I prefer that (I did it in the Babel 7 PR already)

.github/workflows/ci.yaml
keeganwitt: That's a good idea.

SimenB: #336

PR #327, opened 2021-06-23
Upgrade to Babel 7

keeganwitt:
### What

Upgrade to Babel 7.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,16 +35,14 @@
   "license": "MIT",
   "repository": "jest-community/jest-extended",
   "devDependencies": {
+    "@babel/cli": "^7.0.0",
+    "@babel/core": "^7.0.0",
+    "@babel/eslint-parser": "^7.14.7",
+    "@babel/preset-env": "^7.14.7",
     "@types/jest": "^24.0.0",
-    "babel-cli": "^6.26.0",
-    "babel-core": "^6.26.0",
-    "babel-eslint": "^8.0.1",
     "babel-jest": "^24.0.0",
     "babel-jest-assertions": "^0.1.0",
     "babel-plugin-gwt": "^1.0.0",
-    "babel-plugin-transform-async-to-generator": "^6.24.1",
-    "babel-plugin-transform-es2015-modules-commonjs": "^6.26.0",
-    "babel-plugin-transform-object-rest-spread": "^6.26.0",
     "codecov": "^3.0.0",
     "eslint": "^4.9.0",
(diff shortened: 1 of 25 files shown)


package.json
```diff
@@ -35,16 +35,17 @@
   "license": "MIT",
   "repository": "jest-community/jest-extended",
   "devDependencies": {
+    "@babel/cli": "^7.0.0",
+    "@babel/core": "^7.0.0",
+    "@babel/plugin-proposal-object-rest-spread": "^7.0.0",
+    "@babel/plugin-transform-async-to-generator": "^7.0.0",
+    "@babel/plugin-transform-modules-commonjs": "^7.0.0",
```
SimenB, line 42: we don't need any of these when we use the `env` preset

package.json
```diff
@@ -95,11 +96,21 @@
     ]
   },
   "babel": {
+    "presets": [
+      [
+        "@babel/preset-env",
+        {
+          "targets": {
+            "node": "12"
```
SimenB, line 104: this should target whatever is the lowest this library supports - node 8 from the looks of https://github.com/jest-community/jest-extended/blob/437d55ae2e5e3226c9c3cd8bc2317965f9a52a8f/.travis.yml#L3

keeganwitt: Is that still the case? I think that file hasn't been updated in a while. What does Jest support?

package.json
keeganwitt: Are they still needed to be mentioned in the plugins list? Or does that come through present env also? I wasn't totally certain.

```json
"plugins": [
  "@babel/plugin-transform-modules-commonjs",
  "@babel/plugin-proposal-object-rest-spread",
  "@babel/plugin-transform-async-to-generator",
  "module:babel-jest-assertions",
  "gwt",
  "macros"
]
```

SimenB: Remove all mention of them, they're included in the env preset 🙂

keeganwitt: Awesome. Thanks for confirming. Removed.

package.json
SimenB: Jest 26 and 27 supports node 10, but I'm fine with only supporting 12 and later here as 10 is EOL

keeganwitt:
I suspect the failures are because the current version of ESLint doesn't play nice with the upgraded version of `@babel/eslint-parser`. Once we merge #328, I'll rebase and look again at these failures.

keeganwitt:
I went ahead and added the lint fixes to this branch so it can be merged independently.

PR #328, opened 2021-06-23
Update Husky and linting configuration

keeganwitt:
### What

Update Husky and fix linting configuration.

### Why

Linting is not being performed currently by the Git hook.


mattphillips:
I'm not sure I agree with removing the pre-commit hook. Typically I am also against them when they are blocking but in this case all it does is run prettier on the staged changes and is non-blocking.

I agree that it would be good to add a prettier based check to CI also to check consistency as pre-commit hooks are skippable something like: https://prettier.io/docs/en/cli.html#--check

rickhanlonii:
Non-blocking is ok I guess - we should definitely add a check to CI though.

keeganwitt:
Updated to update Husky and the linting configuration so that it runs. Let me know if this isn't what you had in mind.

package.json
```diff
@@ -64,8 +65,8 @@
     "jest-matcher-utils": "^22.0.0"
   },
   "lint-staged": {
-    "*.js": [
-      "yarn prettier",
+    "src/**/*.js": [
+      "yarn run lint:fix",
       "git add"
```
SimenB, line 70: we don't need the `git add` I think

keeganwitt: I wondered about that. It was present before, but I've never had a setup like that.

SimenB (approved):

.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
```
SimenB, line 1: I added to package.json in #334, so you can delete this file

package.json
```diff
@@ -61,8 +62,9 @@
     "jest-matcher-utils": "^22.0.0"
   },
   "lint-staged": {
-    "*.js": "eslint --fix",
-    "*.md": "prettier --write"
+    "src/**/*.js": [
+      "yarn run lint:fix"
+    ]
```
SimenB, line 67: revert these changes

package.json
```diff
@@ -46,7 +47,7 @@
     "eslint-plugin-import": "^2.8.0",
     "eslint-plugin-jest": "^24.3.6",
     "eslint-plugin-prettier": "^3.4.0",
-    "husky": "^0.14.3",
+    "husky": "^6.0.0",
```
SimenB, line 50: update to https://github.com/typicode/husky/releases/tag/v7.0.0?

SimenB (changes requested):

.gitattributes
```diff
@@ -0,0 +1,3 @@
+*.js    text eol=lf
```
keeganwitt, line 1: This probably isn't the best way to fix this, but I tried adding

```json
"endOfLine": "auto",
```

to _package.json_. but I was still getting lint warnings.

rickhanlonii (commented):
LGTM

.gitattributes
rickhanlonii: Would be good to figure out how to not do this, no project I've used with prettier needs it so I'm skeptical that this is the best solution.

rickhanlonii (approved):

SimenB:
note that this didn't actually run - had to `chmod` it: 44b9e6c99110e1e66a39605a4444844a73e9399b

PR #330, opened 2021-06-24
Add Node 12 to engines

keeganwitt:
### What

Document library is targetting Node 12 and newer using `engines`.

### Why

Since GH Actions only test Node 12 and above, and we're talking about 12 and above for Babel settings, we should explicitly document this was our target.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,9 @@
     "jest-get-type": "^22.4.3",
     "jest-matcher-utils": "^22.0.0"
   },
+  "engines": {
+    "node": ">=12"
+  },
   "lint-staged": {
     "*.js": [
       "yarn prettier",
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "jest-matcher-utils": "^22.0.0"
   },
   "engines": {
-    "node": ">=12"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -63,6 +63,9 @@
     "jest-get-type": "^22.4.3",
     "jest-matcher-utils": "^22.0.0"
   },
+  "engines": {
+    "node": ">=12"
```
SimenB, line 67:
```suggestion
    "node": "^12.13.0 || ^14.15.0 || >=16.0.0"
```

to avoid the EOL 13 and 15, plus only support LTS versions of 12 and 14

keeganwitt: Why did you choose `12.13` instead of `12.0`? And `14.15` instead of `14.0`?

keeganwitt: Also, do we want to be that prescriptive? We'd add Node 17 to that list when it comes out? Then remove it when Node 18 comes out?

keeganwitt: Stylistically, it looks like `^` or `>=` can both be used. Do we have a preference?

SimenB: > Why did you choose `12.13` instead of `12.0`? And `14.15` instead of `14.0`?

Those were the first LTS releases of those majors

PR #331, opened 2021-06-27
Make Jest dependencies peer dependencies

keeganwitt:
### What

Moving the Jest dependencies from dev dependencies to peer dependencies.

### Why

Since jest-extended requires Jest dependencies to be hoisted by the project using it (since it's sort of like a plugin for Jest and doesn't have a have and shouldn't have a transitive dependency on Jest itself), we should make that requirement clear by marking Jest as a peer dependency. This should also serve as a resolution for the issue mentioned in #331.

### Notes

1. I'm not quite sure the history behind why the snapshots needed an update. If I'm reading it right, they're semantically equivalent, they just collapsed 2 adjacent `dim` tags into a single tag. I'm guessing this was improved somewhere along the way in Jest and now that the dev dependencies are fully Jest 24, we have a newer Jest version used for tests that include this improvement.
1. Should `@types/jest` be marked as an optional peer since only the jest-extended types would require that to be hoisted?
1. Should `jest-get-type` be marked as an optional peer since only `toBeValid`, `toBeDate`, and `toBeObject` require that to be hoisted?

### Housekeeping

- [x] Unit tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,10 +54,8 @@
     "typescript": "^4.4.3"
   },
   "dependencies": {
-    "expect": "^26.6.2",
     "jest-diff": "^27.2.5",
-    "jest-get-type": "^27.0.6",
-    "jest-matcher-utils": "^27.2.4"
+    "jest-get-type": "^27.0.6"
   },
   "engines": {
     "node": "^12.13.0 || ^14.15.0 || >=16.0.0"
--- a/package.json
+++ b/package.json
@@ -60,6 +60,14 @@
   "engines": {
     "node": "^12.13.0 || ^14.15.0 || >=16.0.0"
   },
+  "peerDependencies": {
+    "@types/jest": ">= 22.0.0"
+  },
+  "peerDependenciesMeta": {
+    "@types/jest": {
+      "optional": true
+    }
+  },
   "lint-staged": {
     "*.js": "eslint --fix",
     "*.md": "prettier --write"
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,10 +30,6 @@ Head over to [here](https://hacktoberfest.digitalocean.com/sign_up/register) to 
   - `index.test.js` - Test suite that uses the new matcher and make sure it passes.
   - `predicate.js` - The function that tests the actual value meets the expected value / behavior.
   - `predicate.test.js` - Tests for the predicate both true/false cases must be covered.
-- [`jest-matchers-utils`](https://github.com/facebook/jest/tree/master/packages/jest-matcher-utils) is being used for syntax highlighting of error messages.
-  - See the Jest docs for an [example usage](https://facebook.github.io/jest/docs/en/expect.html#thisutils)
-- Jest's [`expect`](https://github.com/facebook/jest/tree/master/packages/expect) package is being used to access their deep `equals` function.
-  - `import { equals } from 'expect/build/jasmineUtils';`
 - Docs for the new matcher should be updated in the API section of the `README.md` to no longer say `Unimplemented`.
 - Type definitions for the new matchers should be added to `types/index.d.ts`.
 
--- a/src/matchers/toBeAfter/index.js
+++ b/src/matchers/toBeAfter/index.js
@@ -1,24 +1,22 @@
-import { matcherHint, printReceived } from 'jest-matcher-utils';
-
 import predicate from './predicate';
 
-const passMessage = (received, after) => () =>
-  matcherHint('.not.toBeAfter', 'received', '') +
+const passMessage = (utils, received, after) => () =>
+  utils.matcherHint('.not.toBeAfter', 'received', '') +
   '\n\n' +
-  `Expected date to be after ${printReceived(after)} but received:\n` +
-  `  ${printReceived(received)}`;
+  `Expected date to be after ${utils.printReceived(after)} but received:\n` +
+  `  ${utils.printReceived(received)}`;
 
-const failMessage = (received, after) => () =>
-  matcherHint('.toBeAfter', 'received', '') +
+const failMessage = (utils, received, after) => () =>
+  utils.matcherHint('.toBeAfter', 'received', '') +
   '\n\n' +
-  `Expected date to be after ${printReceived(after)} but received:\n` +
-  `  ${printReceived(received)}`;
+  `Expected date to be after ${utils.printReceived(after)} but received:\n` +
+  `  ${utils.printReceived(received)}`;
 
 export function toBeAfter(date, after) {
   const pass = predicate(date, after);
   if (pass) {
-    return { pass: true, message: passMessage(date, after) };
+    return { pass: true, message: passMessage(this.utils, date, after) };
   }
 
-  return { pass: false, message: failMessage(date, after) };
+  return { pass: false, message: failMessage(this.utils, date, after) };
 }
--- a/src/matchers/toBeAfterOrEqualTo/index.js
+++ b/src/matchers/toBeAfterOrEqualTo/index.js
@@ -1,24 +1,22 @@
-import { matcherHint, printReceived } from 'jest-matcher-utils';
-
 import predicate from './predicate';
 
-const passMessage = (received, before) => () =>
-  matcherHint('.not.toBeAfterOrEqualTo', 'received', '') +
+const passMessage = (utils, received, before) => () =>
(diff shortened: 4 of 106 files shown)


package.json
```diff
@@ -53,12 +53,14 @@
...
+    "jest-get-type": "24.0.0",
     "jest-watch-typeahead": "^0.2.0",
     "lint-staged": "^6.0.0",
     "prettier": "^1.7.4",
     "pretty-format": "^22.1.0"
   },
-  "dependencies": {
+  "peerDependencies": {
+    "@types/jest": "^22.0.0",
     "expect": "^24.1.0",
```
SimenB, line 64: Neither `expect` nor `jest-matcher-utils` should need to be deps - what are they used for?

SimenB (commented):
I don't think making them peers are correct - it'll work with different versions of jest regardless of version mismatch. And either way I think you'll only need `get-types`, I don't think the others are used

package.json
keeganwitt: For `expect`, several predicates have `import { equals } from '../../utils';`, which in turn has
```
import { equals } from 'expect/build/jasmineUtils';
...
export { equals };
```
```
$ grep -l "from '../../utils'" src/matchers/*/*.js
src/matchers/toBeArrayOfSize/index.js
src/matchers/toBeEmpty/predicate.js
src/matchers/toBeOneOf/predicate.js
src/matchers/toContainAllEntries/predicate.js
src/matchers/toContainAllKeys/predicate.js
src/matchers/toContainAllValues/predicate.js
src/matchers/toContainAnyEntries/predicate.js
src/matchers/toContainAnyValues/predicate.js
src/matchers/toContainEntries/predicate.js
src/matchers/toContainEntry/predicate.js
src/matchers/toContainValue/predicate.js
src/matchers/toContainValues/predicate.js
src/matchers/toIncludeAllMembers/predicate.js
src/matchers/toIncludeAnyMembers/predicate.js
src/matchers/toIncludeSameMembers/predicate.js
```
I know there was some talk about whether this was needed, but I didn't honestly understand why.

For `jest-matcher-utils`, most matchers import that
```
$ grep -l "from 'jest-matcher-utils'" src/matchers/*/*.js
src/matchers/toBeAfter/index.js
src/matchers/toBeArray/index.js
src/matchers/toBeArrayOfSize/index.js
src/matchers/toBeBefore/index.js
src/matchers/toBeBoolean/index.js
src/matchers/toBeDate/index.js
src/matchers/toBeEmpty/index.js
src/matchers/toBeEven/index.js
src/matchers/toBeExtensible/index.js
src/matchers/toBeFalse/index.js
src/matchers/toBeFinite/index.js
src/matchers/toBeFrozen/index.js
src/matchers/toBeFunction/index.js
src/matchers/toBeHexadecimal/index.js
src/matchers/toBeNaN/index.js
src/matchers/toBeNegative/index.js
src/matchers/toBeNil/index.js
src/matchers/toBeNumber/index.js
src/matchers/toBeObject/index.js
src/matchers/toBeOdd/index.js
src/matchers/toBeOneOf/index.js
src/matchers/toBePositive/index.js
src/matchers/toBeSealed/index.js
src/matchers/toBeString/index.js
src/matchers/toBeTrue/index.js
src/matchers/toBeValidDate/index.js
src/matchers/toBeWithin/index.js
src/matchers/toContainAllEntries/index.js
src/matchers/toContainAllKeys/index.js
src/matchers/toContainAllValues/index.js
src/matchers/toContainAnyEntries/index.js
src/matchers/toContainAnyKeys/index.js
src/matchers/toContainAnyValues/index.js
src/matchers/toContainEntries/index.js
src/matchers/toContainEntry/index.js
src/matchers/toContainKey/index.js
src/matchers/toContainKeys/index.js
src/matchers/toContainValue/index.js
src/matchers/toContainValues/index.js
src/matchers/toEndWith/index.js
src/matchers/toEqualCaseInsensitive/index.js
src/matchers/toHaveBeenCalledAfter/index.js
src/matchers/toHaveBeenCalledBefore/index.js
src/matchers/toInclude/index.js
src/matchers/toIncludeAllMembers/index.js
src/matchers/toIncludeAnyMembers/index.js
src/matchers/toIncludeMultiple/index.js
src/matchers/toIncludeRepeated/index.js
src/matchers/toIncludeSameMembers/index.js
src/matchers/toReject/index.js
src/matchers/toResolve/index.js
src/matchers/toSatisfy/index.js
src/matchers/toSatisfyAll/index.js
src/matchers/toStartWith/index.js
src/matchers/toThrowWithMessage/index.js
```

Correct me if I've misunderstood something here.

SimenB: All matchers are injected with `this.equals` (and `this.utils`): https://jestjs.io/docs/expect#thisequalsa-b

keeganwitt:
> I don't think making them peers are correct - it'll work with different versions of jest regardless of version mismatch. And either way I think you'll only need `get-types`, I don't think the others are used

What do you mean? That we shouldn't have _any_ peer dependencies (i.e. keep them as dependencies that will be transitively pulled into projects)? Or that the versions that I currently have in the peer dependencies aren't right?

I do think we still need `@types/jest`, since `types/index.d.ts` references that. Unless there's something that injects that also.

SimenB:
`@types/jest` should be an optional peer dep

SimenB:
A separate PR removing the dependencies on `expect` and `jest-matcher-utils` would be nice, then this PR will be smaller after that

keeganwitt:
But removing jest-matcher-utils would jest or dependency to be added by the project using jest-extended to get the injections you mentioned, right? And those aren't currently peer dependencies... Just didn't want to leave it in an inconsistent state between the 2 PRs.

SimenB:
>But removing jest-matcher-utils would jest or dependency to be added by the project using jest-extended to get the injections you mentioned, right?

I don't follow this sentence, sorry 😅 

Maybe this is an answer anyways: When people use `expect.extend`, then `expect` will make sure to inject `this.equals`, `this.utils` etc into any custom matchers - there is no coupling to Jest directly. I don't see any reason why this project would need a (peer) dep for either of those deps. The exception is `jest-get-type`, but imo that can just be a regular dependency. `@types/jest` should be an optional peer dep probably

keeganwitt:
I reworded to be clearer, but I'll try again here.

What I don't understand is if we just delete jest-matcher-utils as a dependency, how does that get hoisted for the person using jest-extended?  It would seem like you're counting on them getting it from somewhere else, but if you've just done the delete alone and not added any peer dependency there's nothing explicitly stating they need to depend on anything else.

Now odds are basically 100% that they will depend on jest, and this will cause that dependency to be hoisted -- but shouldn't that be explicit in the dependencies (i.e. as a peer dependency)?

SimenB:
We don't need anything to be hoisted.

```js
expect.extend({
  myMatcher() {
    console.log(typeof this.equals) // function
    console.log(typeof this.utils.printExpected) // function
  }
})
```

i.e. there is never any need to import the utils, `expect` injects them into the matchers at runtime. No hoisting is needed as `expect` will have the dependency. Where people get `expect` from is up to them - either a direct dependency, relying on Jest injecting it globally or importing from `@jest/globals` are all options, but a library only exposing matchers (like this one) doesn't have to care since it doesn't actually import anything, just exporting functions that expect to be called with the correct context

SimenB:
An example of how this is not needed can be seen in https://github.com/testing-library/jest-dom/pull/250 if that helps clarify the point I'm trying to make 🙂

keeganwitt:
I think I mis-used the term "hoist".

Should we be replacing the imports `from 'jest-matcher-utils'` to use `this.utils` instead? Or am I misunderstanding again?

SimenB:
> Should we be replacing the imports `from 'jest-matcher-utils'` to use `this.utils` instead? Or am I misunderstanding again?

Yeah, that's what I want us to do 👍 same for the `equals` import

keeganwitt:
> > Should we be replacing the imports `from 'jest-matcher-utils'` to use `this.utils` instead? Or am I misunderstanding again?
> 
> Yeah, that's what I want us to do 👍 same for the `equals` import

Gotcha, thank you. Now what you said makes sense :P

keeganwitt:
Made a draft until the dependency refactor mentioned is complete.

keeganwitt:
Created #332 to address previous conversation here.

keeganwitt:
Now that we're deleting expect and jest-matcher-utils, I think there's nothing left to do in this MR. Making `@types/jest` an optional peer dependency was discussed (and dismissed) already in #313.

PR #332, opened 2021-07-01
Delete expect and jest-matcher-utils dependencies

keeganwitt:
### What

Don't depend on expect and jest-matcher-utils directly.

### Why

utils are already injected into all the matchers.

### Housekeeping

- [x] Unit tests

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,10 +30,6 @@ Head over to [here](https://hacktoberfest.digitalocean.com/sign_up/register) to 
   - `index.test.js` - Test suite that uses the new matcher and make sure it passes.
   - `predicate.js` - The function that tests the actual value meets the expected value / behavior.
   - `predicate.test.js` - Tests for the predicate both true/false cases must be covered.
-- [`jest-matchers-utils`](https://github.com/facebook/jest/tree/master/packages/jest-matcher-utils) is being used for syntax highlighting of error messages.
-  - See the Jest docs for an [example usage](https://facebook.github.io/jest/docs/en/expect.html#thisutils)
-- Jest's [`expect`](https://github.com/facebook/jest/tree/master/packages/expect) package is being used to access their deep `equals` function.
-  - `import { equals } from 'expect/build/jasmineUtils';`
 - Docs for the new matcher should be updated in the API section of the `README.md` to no longer say `Unimplemented`.
 - Type definitions for the new matchers should be added to `types/index.d.ts`.
 
--- a/src/matchers/toBeAfter/index.js
+++ b/src/matchers/toBeAfter/index.js
@@ -1,24 +1,22 @@
-import { matcherHint, printReceived } from 'jest-matcher-utils';
-
 import predicate from './predicate';
 
-const passMessage = (received, after) => () =>
-  matcherHint('.not.toBeAfter', 'received', '') +
+const passMessage = (utils, received, after) => () =>
+  utils.matcherHint('.not.toBeAfter', 'received', '') +
   '\n\n' +
-  `Expected date to be after ${printReceived(after)} but received:\n` +
-  `  ${printReceived(received)}`;
+  `Expected date to be after ${utils.printReceived(after)} but received:\n` +
+  `  ${utils.printReceived(received)}`;
 
-const failMessage = (received, after) => () =>
-  matcherHint('.toBeAfter', 'received', '') +
+const failMessage = (utils, received, after) => () =>
+  utils.matcherHint('.toBeAfter', 'received', '') +
   '\n\n' +
-  `Expected date to be after ${printReceived(after)} but received:\n` +
-  `  ${printReceived(received)}`;
+  `Expected date to be after ${utils.printReceived(after)} but received:\n` +
+  `  ${utils.printReceived(received)}`;
 
 export function toBeAfter(date, after) {
   const pass = predicate(date, after);
   if (pass) {
-    return { pass: true, message: passMessage(date, after) };
+    return { pass: true, message: passMessage(this.utils, date, after) };
   }
 
-  return { pass: false, message: failMessage(date, after) };
+  return { pass: false, message: failMessage(this.utils, date, after) };
 }
--- a/package.json
+++ b/package.json
@@ -54,10 +54,8 @@
     "typescript": "^4.4.3"
   },
   "dependencies": {
-    "expect": "^26.6.2",
     "jest-diff": "^27.2.5",
-    "jest-get-type": "^27.0.6",
-    "jest-matcher-utils": "^27.2.4"
+    "jest-get-type": "^27.0.6"
   },
   "engines": {
     "node": "^12.13.0 || ^14.15.0 || >=16.0.0"
--- a/src/matchers/toBeAfterOrEqualTo/index.js
+++ b/src/matchers/toBeAfterOrEqualTo/index.js
@@ -1,24 +1,22 @@
-import { matcherHint, printReceived } from 'jest-matcher-utils';
-
 import predicate from './predicate';
 
-const passMessage = (received, before) => () =>
-  matcherHint('.not.toBeAfterOrEqualTo', 'received', '') +
+const passMessage = (utils, received, before) => () =>
+  utils.matcherHint('.not.toBeAfterOrEqualTo', 'received', '') +
   '\n\n' +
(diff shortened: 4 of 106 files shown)


src/matchers/toBeAfter/index.js
```diff
@@ -1,26 +1,26 @@
...
 
-const failMessage = (received, after) => () =>
-  matcherHint('.toBeAfter', 'received', '') +
+const failMessage = (received, after, context) => () =>
+  context.utils.matcherHint('.toBeAfter', 'received', '') +
   '\n\n' +
-  `Expected date to be after ${printReceived(after)} but received:\n` +
-  `  ${printReceived(received)}`;
-
-export default {
```
SimenB, line 17: keep it inside the `export` to reduce the diff a bit?

```js
export default {
  toBeAfter(date, after) {
    // blahblah
  }
}
```

src/matchers/toBeAfter/index.js
```diff
@@ -1,26 +1,26 @@
...
-      return { pass: true, message: passMessage(date, after) };
-    }
+  `Expected date to be after ${context.utils.printReceived(after)} but received:\n` +
+  `  ${context.utils.printReceived(received)}`;
 
-    return { pass: false, message: failMessage(date, after) };
+function toBeAfter(date, after) {
+  const pass = predicate(date, after);
+  if (pass) {
+    return { pass: true, message: passMessage(date, after, this) };
```
SimenB, line 18:
```suggestion
    return { pass: true, message: passMessage(date, after, this.utils) };
```

src/matchers/toBeAfter/index.js
```diff
@@ -1,26 +1,26 @@
...
+  context.utils.matcherHint('.not.toBeAfter', 'received', '') +
   '\n\n' +
-  `Expected date to be after ${printReceived(after)} but received:\n` +
-  `  ${printReceived(received)}`;
+  `Expected date to be after ${context.utils.printReceived(after)} but received:\n` +
+  `  ${context.utils.printReceived(received)}`;
 
-const failMessage = (received, after) => () =>
-  matcherHint('.toBeAfter', 'received', '') +
+const failMessage = (received, after, context) => () =>
```
SimenB, line 9:
```suggestion
const failMessage = (received, after, { matcherHint, printReceived }) => () =>
```

etc? again, to remove diff

SimenB (approved):
looks great! I have a couple of suggestions to reduce the diff, but the migration itself LGTM. I only commented in the first file, but the comments apply to them all

SimenB:
https://github.com/jest-community/jest-extended/blob/0013709fd2ef28a854fbfff7482221b7e692fc46/src/utils/index.js#L1 remove this?

src/matchers/toBeAfter/index.js
keeganwitt: OK. The one that's a little goofy is _toResolve/index.js_, since it's `async`. Let me know if you can think of a better syntax.

```
export default {
    toResolve: async function toResolve(promise) {
````

**Comment by keeganwitt** [comment]
> https://github.com/jest-community/jest-extended/blob/0013709fd2ef28a854fbfff7482221b7e692fc46/src/utils/index.js#L1
> 
> remove this?

Where does the `equals` come from instead? This is outside a matcher, so nothing is injected, right?

File: src/matchers/toBeAfter/index.js (hunk 2 above)
Comment (keeganwitt) [reply src/matchers/toBeAfter/index.js]: This might be a nice change, would access a bit more directly.

File: src/matchers/toBeAfter/index.js (hunk 3 above)
Comment (keeganwitt) [reply src/matchers/toBeAfter/index.js:9]: It would reduce diff, but I thought it was kinda nice that the matchers all followed the same pattern of all using `context` (although will now be `utils` when I make that change).

**Comment by SimenB** [comment]
yeah, it cannot be imported like this, the matcher must inject it into the helper functions

File: src/matchers/toBeAfter/index.js (hunk 1 above)
Comment (SimenB) [reply src/matchers/toBeAfter/index.js:17]:
```js
export default {
    async toResolve(promise) {},
};
```

this should work?

keeganwitt: I'd tried that, but then tests fail. I don't understand why. Here's one for example.

```
 FAIL  src/matchers/toResolve/index.test.js
  ● .toResolve › passes when passed a promise that resolved

    TypeError: Cannot read property 'utils' of undefined

      11 |     const pass = await predicate(promise);
      12 |     if (pass) {
    > 13 |       return { pass: true, message: passMessage(this.utils) };
         |                                                      ^
      14 |     }
      15 |     return { pass: false, message: failMessage(this.utils) };
      16 |   }
```

keeganwitt:
> yeah, it cannot be imported like this, the matcher must inject it into the helper functions

Got it. Made that change.

src/matchers/toBeAfter/index.js
SimenB: that... sounds weird. Maybe a babel bug (i.e. async to generator transform messes up caller context)? We should land #327 and retry it, and if it's still an issue I guess we can use the longer form 👍

SimenB (commented):
I'll rebase this since I've introduced a conflict in pretty much all files 🙈

src/matchers/toBeAfter/index.js
SimenB: Yep, it works with the latest babel stuff, so you can apply the shorthand to `async` as well 🙂

src/matchers/index.test.js
```diff
@@ -20,25 +20,23 @@ describe('negative matcher', () => {
 describe('asymmetric matchers', () => {
   describe('positive matchers', () => {
     test('.toSatisfy', () => {
-      expect(4).toEqual(expect.toSatisfy(num => num === 4));
+      expect(4).toSatisfy(num => num === 4);
```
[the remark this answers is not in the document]
keeganwitt: I was getting ` Expected at least one assertion to be called but received none.` for some reason. But I think what I changed it to is an equivalent test.

CONTRIBUTING.md
```diff
@@ -33,7 +33,7 @@ Head over to [here](https://hacktoberfest.digitalocean.com/sign_up/register) to
  - [`jest-matchers-utils`](https://github.com/facebook/jest/tree/master/packages/jest-matcher-utils) is being used for syntax highlighting of error messages.
```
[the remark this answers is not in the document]
keeganwitt: Yea, and the bit about `equals` below it too I suppose.

src/matchers/index.test.js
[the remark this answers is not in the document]
SimenB: no, "asymmetric matcher" are e.g. `expect.toSatisfy`, so this is not equivalent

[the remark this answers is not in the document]
keeganwitt: Hmm. These fail at the moment.
> Expected one assertion to be called but received zero assertion calls.

SimenB:
Seems this failed CI, could you rebase to retrigger?

keeganwitt:
> Seems this failed CI, could you rebase to retrigger?

Rebased, but the asymmetric matchers still aren't working.

src/matchers/index.test.js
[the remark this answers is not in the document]
SimenB: Oh, missed this. That seems like a bug in Jest, asymmetric matchers should also receive a `this`... Wanna open up a bug report?

SimenB:
Right, so the failure is (imo) a bug in Jest. I've opened up a PR over there (https://github.com/facebook/jest/pull/11926), but unfortunately that means we can't really land this PR without breaking everybody using the matchers as asymmetric ones.

rickhanlonii:
Could we release a new major of jest-extended instead?

SimenB:
That's easiest, yeah 🙂

keeganwitt:
Jest versions older than v27.2.5 fail because they lack the [asymmetric matchers fix](https://github.com/facebook/jest/pull/11926).

mattphillips (approved):

SimenB:
> Jest versions older than v27.2.5 fail because they lack the [asymmetric matchers fix](https://github.com/facebook/jest/pull/11926).

Yep, so we need to drop those versions from CI and update the peer dependency range before landing

keeganwitt:
Replaced by #405 (because I'm no longer able to push to my fork after having deleted it).

PR #333, opened 2021-07-01
Don't invoke expect.extend

keeganwitt:
### What

Don't extend expect with jest-extended matchers automatically. Let the user choose which to include with Jest in the scope that's appropriate.

### Why

This allows the user more control. Also the existing extend used `global.expect`, which would not be the case when `injectGlobals=false`.

### Housekeeping

- [ ] Unit tests


SimenB:
We should keep the index file, but only export all matchers

keeganwitt:
👍 I started thinking that's what I should do.

As far as the documentation, the extend they need to do now would replace the setupFilesAfterEnv, right?

SimenB:
It should happen in `setupFilesAfterEnv` yes

README.md
```diff
@@ -122,33 +122,46 @@ yarn add -D jest-extended
...
+```javascript
+// ./testSetup.js
+
+// add all jest-extended matchers
+import matchers from 'jest-extended';
+expect.extend(matchers);
+
+// or just add specific matchers
+import { toBeArray, toBeSealed } from 'jest-extended';
+expect.extend(toBeArray, toBeSealed);
```
SimenB, line 134:
```suggestion
expect.extend({ toBeArray, toBeSealed });
```

src/index.js
```diff
@@ -1,15 +1,5 @@
...
-} else {
-  /* eslint-disable no-console */
-  console.error(
-    "Unable to find Jest's global expect." +
-      '\nPlease check you have added jest-extended correctly to your jest configuration.' +
-      '\nSee https://github.com/jest-community/jest-extended#setup for help.'
-  );
-  /* eslint-enable no-console */
-}
+export default {
```
SimenB, line 3: probably `export * from './matchers'`

SimenB (approved):
👍

rickhanlonii:
I'm worried about this being a breaking change. Is there a way we can make this opt-in instead, at least short term? If we want to switch the default, we could add a warning to switch to the new version for at least a major before cutting it all the way over.

SimenB:
This is breaking yes, bit there are already breaking changes on master (dropping old node versions)

rickhanlonii:
I'm more worried about the migration path than just being a breaking change. What if we allowed for something like:

``` json
"jest": {
  "setupFilesAfterEnv": ["jest-extended/all"]
}
```

README.md
```diff
@@ -122,33 +122,46 @@ yarn add -D jest-extended
 
 ## Setup
 
+```javascript
+// ./testSetup.js
+
+// add all jest-extended matchers
+import matchers from 'jest-extended';
```
rickhanlonii, line 129: Should this be 

```js
import * as matchers from 'jest-extended';
```

SimenB:
`["jest-extended/all"]` could work (possibly `["jest-extended/global"]` to indicate what will be used?)

rickhanlonii:
Yeah, either works. That may just be a nice compromise for folks who don't already have a custom setup env and don't want to bother setting one up just for this.

keeganwitt:
Made changes I think as discussed.

I also wasn't sure how we wanted to land this and with what other changes it might be shipped.

src/all/index.js
```diff
@@ -0,0 +1,15 @@
+import matchers from '../matchers';
+
+const jestExpect = global.expect;
+
+if (jestExpect !== undefined) {
+  jestExpect.extend(matchers);
+} else {
+  /* eslint-disable no-console */
+  console.error(
```
SimenB, line 9: we should `throw` since this file is an explicit opt-in

SimenB (approved):

README.md
```diff
@@ -125,33 +125,62 @@ yarn add -D jest-extended
 
 ## Setup
 
+```javascript
+// ./testSetup.js
+
+// add all jest-extended matchers
+import * as matchers from 'jest-extended';
```
SimenB, line 132:
```suggestion
import matchers from 'jest-extended';
```

Looking at https://github.com/jest-community/jest-extended/blob/a0d7d65e997555f07c55b477c4b2ad13ae572b97/src/matchers/index.js#L5 it's a default export

SimenB: it _should_ be named though - thoughts on ditching `import-all.macro` and having explicit named exports?

keeganwitt: Ah. That's what I had originally but changed it because of an earlier comment, since I hadn't tested it.

I had wondered if that'd be better actually, though it would be a bit more verbose. Now that I think of it, the destructuring I added to the Readme might not work without named exports.

keeganwitt: Were you thinking of not using a default export in the index of each of matcher?

SimenB: I have no opinions on that, but `matchers/index.js` should explicitly import all files rather than using `import-all`

SimenB:
@keeganwitt I think we can make a new release after we solve this one

SimenB:
@keeganwitt see https://github.com/jest-community/jest-extended/pull/332#issuecomment-933588737.

I'm fine with only support v27, but we should first make a release of everything we have so people on older versions that 27 can use it

PR #340, opened 2021-09-02
docs(typescript): add triple slash directive

waynevanson:
The import statement did not work for me, but this did.

Here is the file I have that now works.

```ts
/// <reference types="jest-extended" />

/* eslint-disable @typescript-eslint/no-explicit-any */
declare module "all:part:@sanity/base/schema-type" {
  import { BaseSchemaType } from "@sanity/types"
  const anyArray: BaseSchemaType[]

  export default anyArray
}

declare var GRAPHQL_ENDPOINT: string
```

### Housekeeping

- [ ] Unit tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,12 @@ If your editor does not recognise the custom `jest-extended` matchers, add a `gl
 import 'jest-extended';
 ```
 
+If the above import syntax does not work, replace it with the following:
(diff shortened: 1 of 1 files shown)


SimenB:
Hmm, that seems better... @orta which is preferred?

PR #375, opened 2021-10-28
fix(toIncludeAllPartialMembers): includes all member entries

idan-at:
### What

`toIncludeAllPartialMembers` does not check **ALL** members exists, but rather it checks **ANY**.

### Why

According to [the docs](https://github.com/jest-community/jest-extended#toincludeallpartialmembersmembers), toIncludeAllPartialMembers checks if an array contains **all** of the same partial members of a given set.

### Notes
- I noticed repeating code on the `toContainEntries` matchers, so reused existing predicates.

### Housekeeping

- [x] Unit tests


SimenB (approved):
Nice, thanks!

PR #376, opened 2021-10-28
Add toPartiallyContain

### LINKED ISSUE #374: Feature Request: toPartiallyContain
**Feature Request: .toPartiallyContain()**

**Description:**
Checks whether an object is partially contained in a given array.

**Usage example:**
```js
const object = { a: 1, b: 2};
expect([object]).toPartiallyContain({ a: 1 });
```

The example above is shallow, I know. but I think the use case is pretty common - when we have an item, but we don't know / don't want to list all of its properties.

At the moment, jest supports this by using `expect([object]).toEqual(expect.objectContaining({ a: 1}))`, which IMO is cumbersome, and it makes it harder for the reader to understand the expectation.

**Possible solution:**
The logic already exists on `toContainAllEntries`, so it should look something like:

```js
export default (array, item) =>
  Array.isArray(array) &&
  lodash.isPlainObject(item) &&
  array.some(value => toContainAllEntries(value, object.entries(item));
```

idan-at:
### What

Closes #374

### Housekeeping

- [x] Unit tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ If you've come here to help contribute - Thanks! Take a look at the [contributin
     - [.toIncludeAllPartialMembers([members])](#toincludeallpartialmembersmembers)
     - [.toIncludeAnyMembers([members])](#toincludeanymembersmembers)
     - [.toIncludeSameMembers([members])](#toincludesamemembersmembers)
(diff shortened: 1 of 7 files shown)


SimenB (approved):
thanks!

Issue #385, opened 2021-11-13
Define matchers on InverseAsymmetricMatchers for expect.not


dharkness:
The matchers are not visible in auto-completion when used with `expect.not`. They _work,_ but WebStorm reports them as a problem.

```
const value = {
  name: 'foobar',
};
expect(value).toEqual({
  name: expect.not.toStartWith('bar')
                   ^^^^^^^^^^^
                   Unresolved function or method
});
```

The simplest way to expose them is by declaring the asymmetric forms in a new interface, `AsymmetricMatchers`, and having both `Expect` and `InverseAsymmetricMatchers` extend it. However, this may cause issues with TypeScript projects. In that case, I think we'd have to declare the matchers a third time.

Here's how I do it for my own custom matchers:

```
interface AsymmetricMatchers {
  ... custom matchers ...
}

interface Expect extends AsymmetricMatchers {}
interface InverseAsymmetricMatchers extends AsymmetricMatchers {}
```

PR #386, opened 2021-11-13
Declare matchers for use with expect.not.MATCHER (#385)

dharkness:
### What

Move asymmetric matcher declarations to a new `AsymmetricMatchers` interface and extend that to add matchers to `Expect` and `InverseAsymmetricMatchers`.

### Why

This provides autocompletion when using `expect.not`, for example:

```
const value = {
  name: 'foobar',
};
expect(value).toEqual({
  name: expect.not.toStartWith('bar')
});
```

### Notes

This may not play nicely with actual TypeScript projects, but I do not have experience with that to verify. I didn't add any unit tests since I didn't change any code. TypeScript tests would be very useful for the project as a whole.

### Housekeeping

- [ ] Unit tests

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -845,4 +845,8 @@ declare namespace jest {
      */
     toEqualIgnoringWhitespace(string: string): any;
   }
+
+  // noinspection JSUnusedGlobalSymbols
+  // eslint-disable-next-line @typescript-eslint/no-empty-interface
+  interface InverseAsymmetricMatchers extends Expect {}
 }


keeganwitt:
To give me a little more context on this, what autocompletion was this helping enable? With TypeScript projects in In VS Code?

rluvaton:
not related to vs code but typescript not compiling as he thinks these matchers can't be negated using `.not.*`

keeganwitt:
I believe the reason this is happening is because of https://github.com/DefinitelyTyped/DefinitelyTyped/blob/658b4ded756f0c98e1a22e9592014c3c6bc4323a/types/jest/index.d.ts#L729.

SimenB:
https://github.com/jest-community/jest-extended/releases/tag/v3.0.2

PR #402, opened 2022-01-23
fix: remove disguised type assertion

igorwessel:
### What

All the matchers that were using generic, weren't taking advantage of the full potential of typescript.

### Why

The vast majority of matchers related to Array, Object and Satisfy.
Doesn't take advantage of typescript's potential to be able to infer types, and because of that, the object matchers that were related to the object's key allowed passing other keys

without refactor:
example_without_refactor

with refactor:
example_with_refactor

I did this refactoring on all Object, Array and Satisfy matchers.
It was just those who used generics.

### Housekeeping

- [x] Unit tests

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -6,16 +6,12 @@ declare namespace jest {
     /**
      * Note: Currently unimplemented
      * Passing assertion
-     *
-     * @param {String} message
      */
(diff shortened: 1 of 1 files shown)


igorwessel:
I realized it's not a good idea to do this due to the .not operators.

PR #405, opened 2022-01-25
feat: delete expect and jest-matcher-utils dependencies

keeganwitt:
### What

Don't depend on expect and jest-matcher-utils directly.

### Why

utils are already injected into all the matchers.

### Housekeeping

- [x] Unit tests


README.md
```diff
@@ -138,7 +138,7 @@ yarn add -D jest-extended
 
 ## Setup
 
-Note that `jest-extended` only supports Jest version 24 and newer.
+Note that `jest-extended` only supports Jest version 27 and newer.
```
mattphillips, line 141: It might be worth adding for older versions of Jest use `jest-extended@1.x.x` or something

package.json
```diff
@@ -46,18 +46,16 @@
...
     "pretty-format": "^27.2.4",
     "typescript": "^4.4.3"
   },
   "dependencies": {
-    "expect": "^26.6.2",
     "jest-diff": "^27.2.5",
-    "jest-get-type": "^27.0.6",
-    "jest-matcher-utils": "^27.2.4"
+    "jest-get-type": "^27.0.6"
   },
```
mattphillips, line 59: Do we want to add `jest` as a `peerDependency` too? cc/ @SimenB do you know which patch the fix went in for asymmetric matchers?

keeganwitt: I looked for the tags with the commit with the fix, and it was version 27.2.5.

README.md
keeganwitt: Should I bump the version to 2.0.0 in package.json to go with this doc change?

keeganwitt: Also, was there versions that supported Jest older than 24 that I should also mention?

mattphillips: Don't worry about bumping the version to 2.0.0, I'll do that as a tagged commit when releasing it. 

I think jest-extended works with quite a few older versions of Jest so I probably wouldn't specify them in the readme but just say `for older versions of jest use ...`

keeganwitt: I thought about doing this

Note that `jest-extended` only supports Jest version 27 and newer. If you're using an older version of Jest, use the latest jest-extended
version that supports that, as seen below.

| For Jest | Use jest-extended |
| -------- | ----------------- |
| 22       | 0.9.0             |
| 23       | 0.11.1            |
| 24 - 26  | 1.2.0             |
| 27       | \>= 2.0.0         |

package.json
keeganwitt: We'd have to take it out of dev dependencies to do that, right?
```
warning "jest" is already in "devDependencies". Please remove existing entry first before adding it to "peerDependencies".
```

keeganwitt: I guess this is an order of operations thing. If I remove jest, then add as a peer first, then a dev, it doesn't show a warning.

README.md
mattphillips: I like this but am not sure about those version mappings. I vaguely think jest-extended started when Jest was at version 23 (maybe) and suspect it would have worked with older versions of Jest too so I don't know if it is the case the table should just be:

| Jest | Use jest-extended |
| -------- | ----------------- |
| <= 26  | 1.2.0             |
| 27       | \>= 2.0.0         |

@SimenB any ideas?

keeganwitt: Does this answer that question?
```
$ for version in '0.0.1' '0.1.0' '0.2.0' '0.3.0' '0.4.0' '0.4.1' '0.5.0' '0.6.0' '0.6.1' '0.7.0' '0.7.1' '0.7.2' '0.8.0' '0.8.1' '0.9.0' '0.10.0' '0.11.0' '0.11.1' '0.11.2' '0.11.3' '0.11.4' '0.11.5' '1.0.0' '1.1.0' '1.2.0'; do; echo "${version}" && npm info "jest-extended@${version}" dependencies | grep -E '(expect)|(jest-matcher-utils)'; done
0.0.1
{ chalk: '^2.2.0', 'jest-matcher-utils': '^21.2.1' }
0.1.0
{ chalk: '^2.2.0', expect: '^21.2.1', 'jest-matcher-utils': '^21.2.1' }
0.2.0
{ chalk: '^2.2.0', expect: '^21.2.1', 'jest-matcher-utils': '^21.2.1' }
0.3.0
{ chalk: '^2.2.0', expect: '^21.2.1', 'jest-matcher-utils': '^21.2.1' }
0.4.0
{ chalk: '^2.3.0', expect: '^21.2.1', 'jest-matcher-utils': '^21.2.1' }
0.4.1
{ chalk: '^2.3.0', expect: '^21.2.1', 'jest-matcher-utils': '^21.2.1' }
0.5.0
{ chalk: '^2.3.0', expect: '^21.2.1', 'jest-matcher-utils': '^21.2.1' }
0.6.0
{ chalk: '^2.3.0', expect: '^22.1.0', 'jest-matcher-utils': '^22.0.0' }
0.6.1
{ chalk: '^2.3.0', expect: '^22.1.0', 'jest-matcher-utils': '^22.0.0' }
0.7.0
{ chalk: '^2.3.0', expect: '^22.1.0', 'jest-matcher-utils': '^22.0.0' }
0.7.1
{ chalk: '^2.3.0', expect: '^22.1.0', 'jest-matcher-utils': '^22.0.0' }
0.7.2
{ chalk: '^2.3.0', expect: '^22.1.0', 'jest-matcher-utils': '^22.0.0' }
0.8.0
  expect: '^22.1.0',
  'jest-matcher-utils': '^22.0.0'
0.8.1
  expect: '^22.1.0',
  'jest-matcher-utils': '^22.0.0'
0.9.0
  expect: '^22.1.0',
  'jest-matcher-utils': '^22.0.0'
0.10.0
  expect: '^23.6.0',
  'jest-matcher-utils': '^22.0.0'
0.11.0
  expect: '^23.6.0',
  'jest-matcher-utils': '^22.0.0'
0.11.1
  expect: '^23.6.0',
  'jest-matcher-utils': '^22.0.0'
0.11.2
  expect: '^24.1.0',
  'jest-matcher-utils': '^22.0.0'
0.11.3
  expect: '^24.1.0',
  'jest-matcher-utils': '^22.0.0'
0.11.4
  expect: '^24.1.0',
  'jest-matcher-utils': '^22.0.0'
0.11.5
  expect: '^24.1.0',
  'jest-matcher-utils': '^22.0.0'
1.0.0
  expect: '^26.6.2',
  'jest-matcher-utils': '^27.2.4'
1.1.0
  expect: '^26.6.2',
  'jest-matcher-utils': '^27.2.4'
1.2.0
  expect: '^26.6.2',
  'jest-matcher-utils': '^27.2.4'
  ```

keeganwitt: I guess technically because 1.0.0 relies on jest-matcher-utils 27, 1.0.0 already relies on 27? Or is the right way to look at this to only look at expect?

SimenB: I think Matt's approach is the best (i.e. "older use 1.2.0, newer use 2+"). The only thing that really broke compat was supported node versions, not jest versions

package.json
SimenB: yeah, that sounds like a bug in the package manager, perfectly legal (and you essentially _always_ want) to have it in both stanzas

README.md
keeganwitt: OK. If it's just those two to choose from, there's no need for a table. I'll just go back to wording similar to what I originally had.

SimenB (approved):
#yolo

README.md
SimenB: I meant jest version, yes 😀 `engines` is just warning, but unchanged? unless I'm missing something

SimenB:
https://github.com/jest-community/jest-extended/releases/tag/v2.0.0

Herculean PR @keeganwitt, thanks for persisting! 💪

PR #407, opened 2022-01-26
Add global typescript setup example

### LINKED ISSUE #367: Typescript error when importing jest-extended
**Bug**

jest-extended version: 1.0.0

.ts file:
```ts
import * as matchers from "jest-extended";
```

I get error:
File '.../node_modules/jest-extended/types/index.d.ts' is not a module.

Also, after upgrading to 1.0.0 unassigned import does not work any more:
```ts
import "jest-extended";
```

So:
Unassigned import stoped working in version 1.0.0
And wildcard import does not work in typescript because of invalid .d.ts file.

mattphillips:
### What

This PR adds an example project with `jest-extended` setup and working with Typescript

### Why

Fixes #367

I'll add another example for when you may want to use a custom setup that doesn't include all matchers if this is valuable?

### Housekeeping


SimenB (approved):
👍

PR #409, opened 2022-01-29
Remove extraneous all.js file

keeganwitt:
An extra completely unused file was added in a2904bd6b9d1ec215fcf006b00788ea90cf561bf. This removes that file.


keeganwitt:
Actually, this file is published, it's just not inside dist directory.

PR #410, opened 2022-01-29
[WIP] Restructure project

mattphillips:
### What

 - Move `predicate` contents into matcher file
 - Delete `predicate` tests in favour of matcher tests
 - Move matcher tests to `test/matchers/[matcher].js`
 - Move match from `src/matchers/[matcher]/index.js` to `src/matchers/[matcher].js`

### Why

This will simplify the codebase and we be closer to Typescript structured projects as I'd like to migrate to TS once this change is in.

### TODO

Convert all matchers, currently this PR is work in progress

### Housekeeping

- [ ] Unit tests


igorwessel:
If you need any help either here  or in the ts part, I'm available.

mattphillips:
@igorwessel thanks for the offer I think this PR is sorted now but will deffo take you up on the TS offer. I'll also try to go over the TS issue you raised in the next few days when I get some time 😊

CONTRIBUTING.md
```diff
@@ -12,10 +12,6 @@ Thanks for being willing to contribute!
 3. `$ yarn test` to validate you've got it working
 4. Create a branch for your PR
 
-## Here as part of Hacktoberfest?
-
-Head over to [here](https://hacktoberfest.digitalocean.com/sign_up/register) to signup if you haven't already
```
keeganwitt, line 17: Are you no longer interested in participating in Hacktoberfest?

src/matchers/toBeAfterOrEqualTo/predicate.test.js
```diff
@@ -1,18 +0,0 @@
-import predicate from './predicate';
-
-const EARLIER = new Date('01/09/2019');
```
keeganwitt, line 3: _toBeAfterOrEqualTo/index.test.js_ has all the dates formatted in ISO 8601. So deleting this file deletes the only tests with this format. Not sure if we care about that.

src/matchers/toThrowWithMessage/predicate.test.js
```diff
@@ -1,89 +0,0 @@
...
-    ${new RangeError('fox')}       | ${RangeError}     | ${/fox/}
-    ${new ReferenceError('jumps')} | ${ReferenceError} | ${/jumps/}
-    ${new SyntaxError('over')}     | ${SyntaxError}    | ${/over/}
-    ${new URIError('the lazy')}    | ${URIError}       | ${/the lazy/}
-  `('returns false when not instance of $type with the correct error message', ({ error, type, errorMessage }) => {
-    expect(predicate(error, type, errorMessage)).toBe(true);
-  });
-});
-
-describe('toThrowWithMessage Predicate for a string message and same type of error returns false', () => {
```
keeganwitt, line 33: Does this test exist in _index.test.js_?

src/matchers/toBeBefore/predicate.test.js
```diff
@@ -1,14 +0,0 @@
-import predicate from './predicate';
-
-const EARLIER = new Date('06/02/2018');
```
keeganwitt, line 3: _index.test.js_ has all the dates formatted in ISO 8601. So deleting this file deletes the only tests with this format. Not sure if we care about that.

src/matchers/toBeBetween/predicate.test.js
```diff
@@ -1,15 +0,0 @@
-import predicate from './predicate';
-
-const TESTDATE1 = new Date('01/09/2019');
```
keeganwitt, line 3: _index.test.js_ has all the dates formatted in ISO 8601. So deleting this file deletes the only tests with this format. Not sure if we care about that.

src/matchers/toBeEmpty/predicate.test.js
```diff
@@ -1,74 +0,0 @@
...
-
-    test('When empty array is passed', () => {
-      expect(predicate(equals, [])).toBe(true);
-    });
-
-    test('When empty object is passed', () => {
-      expect(predicate(equals, {})).toBe(true);
-    });
-
-    test('When empty Set is passed', () => {
```
keeganwitt, line 22: Set and map tests aren't in _index.test.js_. Not sure if we care about that.

keeganwitt:
Do we want to continue having JSDoc on the new functions?

package.json
```diff
@@ -78,6 +78,9 @@
     "snapshotSerializers": [
       "pretty-format/build/plugins/ConvertAnsi.js"
     ],
+    "moduleNameMapper": {
+      "src/(.*)": "<rootDir>/src/$1"
+    },
```
SimenB, line 83: why do we want this? seems likely to only cause confusion for most tools

mattphillips: So that we can import from `src` in the test files as `src/` and `test/` are at the same level in the directory instead of having to add `../..` to all imports.

CONTRIBUTING.md
mattphillips: Happy to participate but it's once a year thing so probably doesn't need to be in the contributing file all year round (instead we can just tag issues with the `hacktoberfest` label)

src/matchers/toBeAfterOrEqualTo/predicate.test.js
mattphillips: I don't think we care what format the date is constructed from just purely the difference between two dates meets the matcher (`toBeAfter`/`toBeBefore` etc)

package.json
SimenB: right, but is avoiding `../` worth breaking "go to definition" etc?

I don't really have a strong opinion here (nor ownership 😀) but in the same way removing the `import.all` macro makes things not require specific tooling I don't think configuring the resolver of jest and eslint (and later typescript I assume) keeps the code base as approachable as possible.

(if we want something like this, we should use https://nodejs.org/api/packages.html#subpath-imports or something that has an actual spec behind it instead of ad hoc config. That said, Jest does not support it (yet)).

Again, this is your project, not mine, so feel free to structure it how you want to 🙂

mattphillips: > Again, this is your project, not mine, so feel free to structure it how you want to 🙂 Just putting my opinion out there, because what else is the internet for?

Love this 😆

---

Tbh I setup all of my Typescript projects with the ability to import from `src` to make it easy to do absolute imports and move stuff around. With vscode at least (when using TS) importing from `src` doesn't break `go to definition`, do you know if it does in other editors? 

I kind of see this PR as a precursor to switching to TS so hopefully the tooling will be better then. 

This all probably stems from my Java days and when I work in Scala and being able to treat directories as packages/namespaces regardless of source set 😝

src/matchers/toThrowWithMessage/predicate.test.js
mattphillips: See https://github.com/jest-community/jest-extended/pull/410/commits/dae7ff1ce8f9e9e6f27495a2c64d3de53d83c970

package.json
SimenB: I've been writing Rust lately, so I feel this 😀

src/matchers/toBeEmpty/predicate.test.js
mattphillips: see https://github.com/jest-community/jest-extended/pull/410/commits/06c805168ae4c861e8c90084404ba43a115734eb

keeganwitt:
> Do we want to continue having JSDoc on the new functions?

Actually, only _toContainAnyValues/predicate.ts_ and _toBeArrayOfSize/predicate.ts_ had JSDoc (besides _index.d.ts), so we don't need to worry about that. I just randomly happened to look at one of those two that had it.

PR #420, opened 2022-03-02
Update toHaveBeenCalledAfter to fail if second mock is never called

### LINKED ISSUE #379: Why toHaveBeenCalledAfter does not check that the mock has been called?
Hi. I would expect that a test like this exists for `toHaveBeenCalledAfter`:
Fail if the mock under test has not been called at all.

```typescript
  test('fails when given second mock that has not been called and a first mock that has been called', () => {
    const mock1 = jest.fn();
    const mock2 = jest.fn();
    mock1();
    expect(() => expect(mock2).toHaveBeenCalledAfter(mock1)).toThrowError();
  });
```

It is rather surprising for me that "toHaveBeenCalledAfter" does not imply "toHaveBeenCalled". 

With current behavior, I'm forced to write this:
```typescript
    expect(mock2).toHaveBeenCalled();
    expect(mock2).toHaveBeenCalledAfter(mock1);
```

theryansmee:
Update toHaveBeenCalledAfter to fail if second mock is never called Update tests accordingly

✅ Closes #379 

### What
Update toHaveBeenCalledOnce predicate to ensure tests fail if first call is not made

### Why

- In #379 it is agreed that `toHaveBeenCalledOnce` should also ensure that `toHaveBeenCalled` is also true

### Housekeeping

- [x] Unit tests

### DIFF
--- a/src/matchers/toHaveBeenCalledAfter.js
+++ b/src/matchers/toHaveBeenCalledAfter.js
@@ -37,7 +37,7 @@ export function toHaveBeenCalledAfter(actual, expected) {
 const smallest = ns => ns.reduce((acc, n) => (acc < n ? acc : n));
 
 const predicate = (firstInvocationCallOrder, secondInvocationCallOrder) => {
-  if (firstInvocationCallOrder.length === 0) return true;
+  if (firstInvocationCallOrder.length === 0) return false;
   if (secondInvocationCallOrder.length === 0) return false;
 
   const firstSmallest = smallest(firstInvocationCallOrder);
--- a/src/matchers/toHaveBeenCalledAfter.js
+++ b/src/matchers/toHaveBeenCalledAfter.js
@@ -1,6 +1,6 @@
 import { isJestMockOrSpy } from '../utils';
 
-export function toHaveBeenCalledAfter(actual, expected) {
+export function toHaveBeenCalledAfter(actual, expected, failIfNoFirstInvocation = true) {
   const { printReceived, printExpected, matcherHint } = this.utils;
 
   if (!isJestMockOrSpy(actual)) {
@@ -13,7 +13,7 @@ export function toHaveBeenCalledAfter(actual, expected) {
 
   const firstInvocationCallOrder = actual.mock.invocationCallOrder;
   const secondInvocationCallOrder = expected.mock.invocationCallOrder;
-  const pass = predicate(firstInvocationCallOrder, secondInvocationCallOrder);
+  const pass = predicate(firstInvocationCallOrder, secondInvocationCallOrder, failIfNoFirstInvocation);
 
   const passMessage =
     matcherHint('.not.toHaveBeenCalledAfter') +
@@ -36,8 +36,8 @@ export function toHaveBeenCalledAfter(actual, expected) {
 
 const smallest = ns => ns.reduce((acc, n) => (acc < n ? acc : n));
 
-const predicate = (firstInvocationCallOrder, secondInvocationCallOrder) => {
-  if (firstInvocationCallOrder.length === 0) return false;
+const predicate = (firstInvocationCallOrder, secondInvocationCallOrder, failIfNoFirstInvocation) => {
+  if (firstInvocationCallOrder.length === 0) return !failIfNoFirstInvocation;
   if (secondInvocationCallOrder.length === 0) return false;
 
   const firstSmallest = smallest(firstInvocationCallOrder);
--- a/src/matchers/toHaveBeenCalledBefore.js
+++ b/src/matchers/toHaveBeenCalledBefore.js
@@ -1,6 +1,6 @@
 import { isJestMockOrSpy } from '../utils';
 
-export function toHaveBeenCalledBefore(actual, expected) {
+export function toHaveBeenCalledBefore(actual, expected, failIfNoSecondInvocation = true) {
   const { printReceived, printExpected, matcherHint } = this.utils;
 
   if (!isJestMockOrSpy(actual)) {
@@ -13,7 +13,7 @@ export function toHaveBeenCalledBefore(actual, expected) {
 
   const firstInvocationCallOrder = actual.mock.invocationCallOrder;
   const secondInvocationCallOrder = expected.mock.invocationCallOrder;
-  const pass = predicate(firstInvocationCallOrder, secondInvocationCallOrder);
+  const pass = predicate(firstInvocationCallOrder, secondInvocationCallOrder, failIfNoSecondInvocation);
 
   const passMessage =
     matcherHint('.not.toHaveBeenCalledBefore') +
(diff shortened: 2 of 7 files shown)


src/matchers/toHaveBeenCalledAfter/index.test.js
```diff
@@ -73,12 +74,6 @@ describe('.toHaveBeenCalledAfter', () => {
 });
 
 describe('.not.toHaveBeenCalledAfter', () => {
-  test('fails when given a first mock has not been called', () => {
```
SimenB, line 76: why not this? duplication?

theryansmee: Yeah, jest warned me that it was obsolete because it was a duplicate 😊

SimenB: fancy stuff 😅

keeganwitt:
Do you think we should also change `toBeBefore` to fail if the second mock is never called?

theryansmee:
> Do you think we should also change `toBeBefore` to fail if the second mock is never called?

I have been contemplating the same thing. 

I don't think it's as necessary as the change to `toBeAfter`, but i'm happy to take a look in the interest of keeping the behaviour similar between the two?

keeganwitt:
> > Do you think we should also change `toBeBefore` to fail if the second mock is never called?
> 
> I have been contemplating the same thing.
> 
> I don't think it's as necessary as the change to `toBeAfter`, but i'm happy to take a look in the interest of keeping the behaviour similar between the two?

That's what I was thinking was to keep the behavior similar.  I was on the fence before, but the more I started thinking about it, the more I think it makes sense.  If we don't check that the second call happened, there's no difference between `toHaveBeenCalledBefore` and `toHaveBeenCalled()` in cases where the second call didn't happen.

Another thing I'm thinking about is whether we should make this behavior configurable too, like
```patch
Index: src/matchers/toHaveBeenCalledAfter.js
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/matchers/toHaveBeenCalledAfter.js b/src/matchers/toHaveBeenCalledAfter.js
--- a/src/matchers/toHaveBeenCalledAfter.js	(revision 505974f1ad3a511c6a364c8c09230775cbd955f9)
+++ b/src/matchers/toHaveBeenCalledAfter.js	(date 1646850139144)
@@ -1,6 +1,6 @@
 import { isJestMockOrSpy } from '../utils';
 
-export function toHaveBeenCalledAfter(actual, expected) {
+export function toHaveBeenCalledAfter(actual, expected, failIfNoFirstInvocation) {
   const { printReceived, printExpected, matcherHint } = this.utils;
 
   if (!isJestMockOrSpy(actual)) {
@@ -13,7 +13,7 @@
 
   const firstInvocationCallOrder = actual.mock.invocationCallOrder;
   const secondInvocationCallOrder = expected.mock.invocationCallOrder;
-  const pass = predicate(firstInvocationCallOrder, secondInvocationCallOrder);
+  const pass = predicate(firstInvocationCallOrder, secondInvocationCallOrder, !!failIfNoFirstInvocation);
 
   const passMessage =
     matcherHint('.not.toHaveBeenCalledAfter') +
@@ -36,8 +36,8 @@
 
 const smallest = ns => ns.reduce((acc, n) => (acc < n ? acc : n));
 
-const predicate = (firstInvocationCallOrder, secondInvocationCallOrder) => {
-  if (firstInvocationCallOrder.length === 0) return false;
+const predicate = (firstInvocationCallOrder, secondInvocationCallOrder, failIfNoFirstInvocation) => {
+  if (firstInvocationCallOrder.length === 0) return !failIfNoFirstInvocation;
   if (secondInvocationCallOrder.length === 0) return false;
 
   const firstSmallest = smallest(firstInvocationCallOrder);

```

theryansmee:
> 



> > > Do you think we should also change `toBeBefore` to fail if the second mock is never called?
> > 
> > 
> > I have been contemplating the same thing.
> > I don't think it's as necessary as the change to `toBeAfter`, but i'm happy to take a look in the interest of keeping the behaviour similar between the two?
> 
> That's what I was thinking was to keep the behavior similar. I was on the fence before, but the more I started thinking about it, the more I think it makes sense. If we don't check that the second call happened, there's no difference between `toHaveBeenCalledBefore` and `toHaveBeenCalled()` in cases where the second call didn't happen.
> 
> Another thing I'm thinking about is whether we should make this behavior configurable too, like
> 
> ```diff
> diff --git a/src/matchers/toHaveBeenCalledAfter.js b/src/matchers/toHaveBeenCalledAfter.js
> index 01bb87a..5aa6182 100644
> --- a/src/matchers/toHaveBeenCalledAfter.js
> +++ b/src/matchers/toHaveBeenCalledAfter.js
> @@ -1,6 +1,6 @@
>  import { isJestMockOrSpy } from '../utils';
> 
> -export function toHaveBeenCalledAfter(actual, expected) {
> +export function toHaveBeenCalledAfter(actual, expected, failIfNoFirstInvocation) {
>    const { printReceived, printExpected, matcherHint } = this.utils;
> 
>    if (!isJestMockOrSpy(actual)) {
> @@ -13,7 +13,7 @@ export function toHaveBeenCalledAfter(actual, expected) {
> 
>    const firstInvocationCallOrder = actual.mock.invocationCallOrder;
>    const secondInvocationCallOrder = expected.mock.invocationCallOrder;
> -  const pass = predicate(firstInvocationCallOrder, secondInvocationCallOrder);
> +  const pass = predicate(firstInvocationCallOrder, secondInvocationCallOrder, !!failIfNoFirstInvocation);
> 
>    const passMessage =
>      matcherHint('.not.toHaveBeenCalledAfter') +
> @@ -36,8 +36,8 @@ export function toHaveBeenCalledAfter(actual, expected) {
> 
>  const smallest = ns => ns.reduce((acc, n) => (acc < n ? acc : n));
> 
> -const predicate = (firstInvocationCallOrder, secondInvocationCallOrder) => {
> -  if (firstInvocationCallOrder.length === 0) return false;
> +const predicate = (firstInvocationCallOrder, secondInvocationCallOrder, failIfNoFirstInvocation) => {
> +  if (firstInvocationCallOrder.length === 0) return !failIfNoFirstInvocation;
>    if (secondInvocationCallOrder.length === 0) return false;
> 
>    const firstSmallest = smallest(firstInvocationCallOrder);
> ```

Yeah thats a really good shout!
I also think that `failIfNoFirstInvocation` should default to `true` so the user only passes it if they don't care about both mocks being called? (The only draw back being that this would then be a breaking change as currently tests pass without the second call). 

What are your thoughts?

Also, I have made the requested changes to `toHaveBeenCalledBefore`

theryansmee:
@keeganwitt - When you get a chance, have a look at my last commit and let me know what you think of my implementation of the discussed param

keeganwitt:
> I also think that failIfNoFirstInvocation should default to true so the user only passes it if they don't care about both mocks being called? (The only draw back being that this would then be a breaking change as currently tests pass without the second call).

Yea, that is a better way to do it.  I'm still mulling the overall idea over, particularly the point you raise about it being a breaking change, but I'm thinking this is the right approach.

Issue #138, opened 2018-06-21
New matchers: toHaveBeenCalledOnce and toHaveBeenCalledOnceWith


amercier:
I think it would be nice to add the following matchers for `Mock`:

### toHaveBeenCalledOnce

```js
const mock = jest.fn();
// ...
expect(mock).toHaveBeenCalledOnce();
```

that would be equivalent to:

```js
const mock = jest.fn();
// ...
expect(mock).toHaveBeenCalledTimes(1);
```

### toHaveBeenCalledOnceWith

```js
const mock = jest.fn();
// ...
expect(mock).toHaveBeenCalledOnceWith('foo', 'bar');
```

that would be equivalent to:

```js
const mock = jest.fn();
// ...
expect(mock).toHaveBeenCalledTimes(1);
expect(mock).toHaveBeenLastCalledWith('foo', 'bar');
```

### Rationale

I find myself regularly writing tests with only one call per mock. Especially while writing unit tests, where I prefer writing many small, atomic tests, rather than one large test for all cases.

benjaminkay93:
if nothing else im happy to raise a pr for this

Julian-Jurai:
We have something similar but not as semantic : 

```javascript
const mock = jest.fn();
expect(mock).toHaveBeenNthCalledWith(1, 'foo', 'bar');
```javascript

mattphillips:
Sounds great! Feel free to send a PR with this, just drop a note here if you are working on it :smile:

benjaminkay93:
Ive got a branch locally with a toHaveBeenCalledOnce matcher, it was queit straightforward, however the toHaveBeenCalledOnceWith is causing my brain issues, should we use === equality on objects or == equality on them?

mattphillips:
@benjaminkay93 I think we probably want to do deep equality on them you can use the [`equals`](https://github.com/jest-community/jest-extended/blob/master/src/utils/index.js#L11) util to compare two objects deeply or the [`contains`](https://github.com/jest-community/jest-extended/blob/master/src/utils/index.js#L3) util to deeply check if an array contains an object

benjaminkay93:
Sweet thank you @mattphillips, big help

theryansmee:
@SimenB - I'll take a look at `toHaveBeenCalledOnceWith` 😊

PR #430, opened 2022-03-23
Add toHaveBeenCalledOnceWith matcher

thibautsabot:
### What

Add toHaveBeenCalledOnceWith matcher

### Why

To fix the second part of https://github.com/jest-community/jest-extended/issues/138

### Housekeeping

- [x] Unit tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@ If you've come here to help contribute - Thanks! Take a look at the [contributin
     - [.toHaveBeenCalledBefore()](#tohavebeencalledbefore)
     - [.toHaveBeenCalledAfter()](#tohavebeencalledafter)
     - [.toHaveBeenCalledOnce()](#tohavebeencalledonce)
+    - [.toHaveBeenCalledOnceWith()](#tohavebeencalledoncewith)
   - [Number](#number)
     - [.toBeNumber()](#tobenumber)
     - [.toBeNaN()](#tobenan)
@@ -610,6 +611,20 @@ it('passes only if mock was called exactly once', () => {
   expect(mock).not.toHaveBeenCalled();
   mock();
   expect(mock).toHaveBeenCalledOnce();
+});
+```
+
+#### .toHaveBeenCalledOnceWith()
+
+Use `.toHaveBeenCalledOnceWith` to check if a `Mock` was called exactly one time with the expected value.
+
+```js
+it('passes only if mock was called exactly once with the expected value', () => {
+  const mock = jest.fn();
+
+  expect(mock).not.toHaveBeenCalled();
+  mock('hello');
+  expect(mock).toHaveBeenCalledOnceWith('hello');
 });
 ```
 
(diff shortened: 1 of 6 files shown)


keeganwitt:
I think it'd be more flexible to do this using the existing matchers. For example,
```javascript
const arg = 1;
mock(arg);
expect(mock).toHaveBeenCalledOnce();
expect(mock.mock.calls[0][0]).toEqual(arg);
```

Using this pattern, you can use `toEqual()` or your choice of any matcher instead of only `equals()`.

keeganwitt:
Thank you for the suggestion and the PR (sorry I forgot to say that earlier)!

The fact Jest's own `toHaveBeenCalledWith` uses `equals()` is a good point. Lemme see if I can get another opinion. Our goal is to add matchers that add enough value to be worth adding, and not every conceivable matcher. I'm still getting a feel for where to draw the line.

G-Rath:
I think it makes sense to have this match if we've got `toHaveBeenCalledOnce` as it is a natural complement & it doesn't add that much (I'd actually think we could try and deduplicate the code with the non-`with` matcher, but that doesn't have to be in this PR), _but_ I would expect it to try and use the same logic as `toHaveBeenCalledWith`, whatever that is, since this is meant to be in effect the same matcher just combined with a count checked.

G-Rath (approved):

G-Rath:
I'm trying to find the underlying matcher in `jest` which I think is here: https://github.com/facebook/jest/blob/51fa61964f0b8f4d06db93a014b49b21ec53ea71/packages/expect/src/spyMatchers.ts#L553

I think we should confirm if the checks there are the equivalent to `this.equal`.

Fouzyyyy:
Hi there

I thought about having a util function in our company codebase that is called `toHaveBeenCalledOnceWith` and under the hood, it calls toHaveBeenCalledTimes(1) and toHaveBeenCalledWith()

How is this different from the proposed implementation here?

keeganwitt:
It would do the same thing. So would `toBeCalledTimes(1)` and `toHaveBeenNthCalledWith(1, expected)`.

keeganwitt:
I'm wondering if it'd maybe be better to change this to `toHaveBeenCalledNTimesWith`, where the equivalent of the current implementation would be `toHaveBeenCalledNTimesWith(1, expected)`. Thoughts?

monokrome:
**toHaveBeenCalledNTimesWith** could be a separate matcher, but I'd hate to conflate it with this PR personally. In my personal experience, **toHaveBeenCalledOnceWith** is a much more common use-case.

Also, if we wanted to consider **ToHaveBeenCalledNTimesWith**, we're going to have to have the discussion of how the **expected** argument works. Can I pass multiple arguments as **...expected** for each call? Does it just match every call expecting the same ones? etc. At that point, it seems questionable whether it's easier to use this matcher or we should just manually check each call consecutively.

monokrome:
@thibautsabot FYI it looks like this was approved 😺

keeganwitt:
> I'm trying to find the underlying matcher in `jest` which I think is here: https://github.com/facebook/jest/blob/51fa61964f0b8f4d06db93a014b49b21ec53ea71/packages/expect/src/spyMatchers.ts#L553
> 
> I think we should confirm if the checks there are the equivalent to `this.equal`.

I'm pretty sure those are the same.

keeganwitt:
#518 made the type match the actual implementation. #517 would change the implementation to accept more than just 1 argument.

PR #439, opened 2022-04-25
feat: upgrade to Jest 28 dependencies

SimenB:
should work


SimenB:
Bah, absolutely does not 😅 moving config to js file is probably easiest. I might take a look over the weekend

Issue #472, opened 2022-07-15
[Bug] Type definition of .toThrowWithMessage() is broken


sounisi5011:
**Bug**

- `package` version: [`jest-extended@3.0.1`](https://github.com/jest-community/jest-extended/tree/v3.0.1)
- `node` version: 16.16.0 or 14.19.3 (This bug is caused by a TypeScript type definition and has nothing to do with the Node.js version)
- `pnpm` version: 6.33.0

Relevant code

```ts
expect(() => {
  throw TypeError('hello world');
}).toThrowWithMessage(TypeError, 'hello world');

expect.toThrowWithMessage(TypeError, 'hello world');
```

What you did:

I used TypeScript's `tsc` command to check for the presence of errors caused by type definitions in the test code.

What happened (please provide anything you think will help):

Due to [a change](https://github.com/jest-community/jest-extended/pull/471/files#diff-093ad82a25aee498b11febf1cdcb6546e4d223ffcb49ed69cc275ac27ce0ccceR390) introduced in #471, the constructor function cannot be specified as the first argument of toThrowWithMessage. If tried, the following error will be thrown:

```
error TS2345: Argument of type 'TypeErrorConstructor' is not assignable to parameter of type '(...args: unknown[]) => unknown'.
  Types of parameters 'message' and 'args' are incompatible.
    Type 'unknown' is not assignable to type 'string | undefined'.

3 }).toThrowWithMessage(TypeError, 'hello world');
                        ~~~~~~~~~
```

```
error TS2345: Argument of type 'TypeErrorConstructor' is not assignable to parameter of type '(...args: unknown[]) => unknown'.
  Types of parameters 'message' and 'args' are incompatible.
    Type 'unknown' is not assignable to type 'string | undefined'.

5 expect.toThrowWithMessage(TypeError, 'hello world');
                            ~~~~~~~~~
```

[TS Playground link](https://www.typescriptlang.org/play?ts=4.7.2#code/CYUwxgNghgTiAEYD2A7AzgF3hpAVAFjEgO4DqAlhvgLIhppQDmIAXPABQYCeADqxwDohsRmjYBXFAGsUJFAG0AugEp4AXgB88STLkAaeAFs6DZm0wxyKRvAA+8AEohGAUQAePVZvgA3JOWAAbgBYACgwnAIiMkoaEyYQdlxeEBcYIhgDAHJ8EAgIJHhiJBgIYCzlQKA)

---

[...]

PR #475, opened 2022-07-18
Fix typings of `toThrowWithMessage`, support unconstructable errors

GerkinDev:
### What

This PR fixes typings issues noted in #472, and adds support to unconstructable errors (such as abstract classes or not taking a `message` string as first arg) to be passed as error class.

### Why

See #472. Moreover, `toThrowWithMessage` was unusable with `AssertionError` from the node `assert` module.

### Notes

The matcher first tries to instantiate the `type` with the message (`toString`ed in case it was a regex).
If the instantiation fails, it creates a new `Error` on which it sets the `name` & `message` in order to be correctly printed as expected in the message.

### Housekeeping

- [x] Unit tests

### DIFF
--- a/src/matchers/toThrowWithMessage.js
+++ b/src/matchers/toThrowWithMessage.js
@@ -85,9 +85,19 @@ export function toThrowWithMessage(callbackOrPromiseReturn, type, message) {
   }
 
   const pass = predicate(error, type, message);
+  const messageStr = message.toString();
+  let expectedError;
+  try {
+    expectedError = new type(messageStr);
+  } catch (err) {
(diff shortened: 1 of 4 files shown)


SimenB (approved):
sweet, thanks!

GerkinDev:
Hi @SimenB , do you know when this will be released ? I'm heavily relying on node `assert` and I can't test thrown errors correctly without this.

Thanks !

SimenB:
https://github.com/jest-community/jest-extended/releases/tag/v3.0.2

GerkinDev:
Thanks so much !

PR #477, opened 2022-07-20
Change any to unknown

keeganwitt:
A follow-up to [this conversation](https://github.com/jest-community/jest-extended/pull/292#discussion_r725865339). We need the second argument to be an array, then we can use `unknown` instead of `any`, and this is the best practice.

After starting this change, I realized there were multiple places we could change an `any` to `unknown`.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -51,9 +51,7 @@
       "plugins": ["@typescript-eslint"],
       "extends": ["plugin:@typescript-eslint/recommended"],
       "rules": {
-        "@typescript-eslint/array-type": "error",
-        // should probably enable this, but it's a bit too strict for now
-        "@typescript-eslint/no-explicit-any": "off"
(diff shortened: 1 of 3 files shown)


SimenB:
Remove this as well? https://github.com/jest-community/jest-extended/blob/801c06865defc98bd2ac59a9f938f7f118af45f1/.eslintrc#L56

keeganwitt:
> Remove this as well?
> 
> https://github.com/jest-community/jest-extended/blob/801c06865defc98bd2ac59a9f938f7f118af45f1/.eslintrc#L55-L56

There are 71 `any` usages left still in `index.d.ts`. However, there aren't any elsewhere, so maybe it'd still be a better default and turn off the rule just for this 1 file.

SimenB:
Only ts-file we've got atm, so that makes sense

SimenB (approved):

PR #482, opened 2022-08-05
Add failing element to `satisfyAll` and `satisfyAny` error messages

mattphillips:
### What

Add failing element to `satisfyAll` and `satisfyAny` error messages

### Why

This will give a clearer error message of why the assertion failed

### Notes

Screenshot 2022-08-05 at 13 21 09

### Housekeeping

- [x] Unit tests

### DIFF
--- a/src/matchers/toSatisfyAll.js
+++ b/src/matchers/toSatisfyAll.js
@@ -1,13 +1,18 @@
 export function toSatisfyAll(actual, expected) {
   const { printReceived, printExpected, matcherHint } = this.utils;
 
+  const pass = actual.every(expected);
+  const failingElement = pass ? actual.find(expected) : actual.find(a => !expected(a));
+
   const passMessage =
     matcherHint('.not.toSatisfyAll') +
     '\n\n' +
     'Expected array to not satisfy predicate for all values:\n' +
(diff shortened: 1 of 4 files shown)


src/matchers/toSatisfyAny.js
```diff
@@ -1,23 +1,28 @@
 export function toSatisfyAny(actual, expected) {
   const { printReceived, printExpected, matcherHint } = this.utils;
 
+  const pass = actual.some(expected);
+  const failingElement = pass ? actual.find(expected) : actual.find(a => !expected(a));
```
keeganwitt, line 5: Why return the first non-matching element? In the case of `satisfyAny`, all elements would fail the predicate.

src/matchers/toSatisfyAll.js
```diff
@@ -1,23 +1,28 @@
 export function toSatisfyAll(actual, expected) {
   const { printReceived, printExpected, matcherHint } = this.utils;
 
+  const pass = actual.every(expected);
+  const failingElement = pass ? actual.find(expected) : actual.find(a => !expected(a));
```
keeganwitt, line 5: There could be multiple failing elements, right? I'm wondering if it'd be better to print them all instead of just the first one.

PR #515, opened 2022-10-04
Add `vitest` info to docs

silverwind:
### What

Add docs for `vitest` integration. Reference: https://vitest.dev/api/#expect-extend

### Why

It's generally useful for `vitest` users

### Housekeeping

### DIFF
--- a/website/docs/getting-started/setup.md
+++ b/website/docs/getting-started/setup.md
@@ -45,7 +45,6 @@ To automatically extend `expect` with all matchers, you can use
 ```javascript
 import {expect} from "vitest";
 import * as matchers from "jest-extended";
-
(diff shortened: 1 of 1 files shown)


website/docs/getting-started/setup.md
```diff
@@ -37,3 +37,24 @@ To automatically extend `expect` with all matchers, you can use
...
 ```
+
+## Use with `vitest`
+
+`jest-extended` works with `vitest` because their `expect.extend` API is compatible. In your setup script:
+
+```javascript
+import {expect} from "vitest";
+import * as matchers from "jest-extended";
+
```
silverwind, line 48:
```suggestion
```

PR #520, opened 2022-10-17
Mark jest peerDependency as optional

silverwind:
### What

Mark the jest dependency as optional

### Why

The module works with other runners like vitest as well, in which case, the user will not need to have jest installed to use this package.

### Housekeeping

- [ ] Unit tests


SimenB (approved):
Very reasonable 👍

Issue #544, opened 2022-12-27
Matchers very slow for large objects


overlookmotel:
**Bug**

- `package` version: 3.2.0
- `node` version: 16.18.1
- `npm` version: 8.19.2

Relevant code or config:

```javascript
const obj = {};
for (let i = 0; i < 100000; i++) {
  obj[`_${i}`] = i;
}

it('is an object', () => {
  expect(obj).toBeObject();
});
```

This test takes approx 7 seconds to complete. NB that's 7 secs just to run the expectation, not including building the object.

Reason is that `toBeObject` creates both the pass and fail messages even if the test passes. This includes calling `printReceived()` on the input, and when the input is large, this takes a long time.

Neither message is used when the test passes, so this is unnecessary.

Pretty much all the other matchers are affected by the same issue.

I will submit a PR to fix this.

PR #545, opened 2022-12-27
Create pass/fail error messages only if required

overlookmotel:
### What

Issue: #544

Almost all the matchers eagerly create both `passMessage` and `failMessage` regardless of whether the expectation passes or not. In the common case that the expectation passes, neither error message is ever used, so all this work is unnecessary. If the expectation fails, only one of the pass/fail message is used, so half the work is unnecessary.

This is slowing down all expectations, but the problem becomes very noticeable if the input is a large object, so calculating `printReceived()` is very slow (see #544).

This PR amends all the matchers to lazily calculate the pass/fail error messages only if they're required.

### Why

To make tests run faster!

Note on implementation:

This PR has a large diff because it touches almost all the matchers, but the changes are mostly trivial. In almost every case, I've simply moved the content of `const passMessage` and `const failMessage` into inside the `message` function.

The only matchers with any more complex changes than this are:

* `toBeArrayOfSize` - reformatted for consistency with the rest
* `toEqualIgnoringWhitespace` - create diff only if required, extra test to maintain 100% coverage
* `toHaveBeenCalledOnceWith` - 3 different messages
* `toThrowWithMessage` - create `expectedError` only if required

I hope this info makes it less onerous to review this PR.

### Housekeeping

- [x] Unit tests

### DIFF
--- a/src/matchers/toBeAfter.js
+++ b/src/matchers/toBeAfter.js
@@ -1,18 +1,19 @@
 export function toBeAfter(date, after) {
   const { printReceived, matcherHint } = this.utils;
-  const passMessage =
-    matcherHint('.not.toBeAfter', 'received', '') +
-    '\n\n' +
-    `Expected date to be after ${printReceived(after)} but received:\n` +
-    `  ${printReceived(date)}`;
-
-  const failMessage =
-    matcherHint('.toBeAfter', 'received', '') +
-    '\n\n' +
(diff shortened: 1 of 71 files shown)


SimenB (approved):
thanks!! 🎉

SimenB:
https://github.com/jest-community/jest-extended/releases/tag/v3.2.1

Issue #548, opened 2023-01-02
Vitest TypeScript support for InverseAsymmetricMatchers


rluvaton:
- `package` version: 3.2.1
- `node` version: 18.12.1
- `npm` (or `yarn`) version: 8.19.2

What happened (please provide anything you think will help):

Typescript fail in this
```typescript
test('InverseAsymmetricMatchers', () => {
	expect({
		data: [1, 2]
	}).toEqual({

		// I get here:
		// Property 'toIncludeSameMembers' does not exist on type 'ExpectStatic'
		data: expect.toIncludeSameMembers([2, 1])
	})
});
```

the type failed:
```
Property 'toIncludeSameMembers' does not exist on type 'ExpectStatic'
```

Reproduction repository (if possible):
https://github.com/rluvaton/jest-extended-bug-548

just run:
```
npm install
npm start
```

SimenB:
Happy to take a PR fixing this

rluvaton:
I can try but one thing that can help me is how the regular marchers works with typescript, the types literally have jest in it:
```typescript
declare namespace jest {
```

I would expect something related to the [`expect`](https://www.npmjs.com/package/expect) package...

Great work BTW on all the jest stuff!

SimenB:
Yeah, the types shipped with Jest itself uses `expect` (https://jestjs.io/docs/expect#expectextendmatchers), but _this_ package enhances types from `@types/jest`.

rluvaton:
created a PR

SimenB:
https://github.com/jest-community/jest-extended/releases/tag/v3.2.2

rluvaton:
Happy to contribute to open source, hope to see you again

PR #549, opened 2023-01-05
fix(types): add vitest support for AsymmetricMatchers

rluvaton:
Fix https://github.com/jest-community/jest-extended/issues/548

@SimenB should I update the package version?

### Housekeeping

- [ ] Unit tests

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,13 +1,13 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
-interface CustomMatchers<R = unknown> {
+interface CustomMatchers {
   /**
    * Note: Currently unimplemented
    * Passing assertion
    *
    * @param {String} message
    */
-  pass(message: string): R;
+  pass(message: string): void;
 
   /**
    * Note: Currently unimplemented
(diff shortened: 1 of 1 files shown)


types/index.d.ts
```diff
@@ -876,3 +879,8 @@ declare namespace jest {
   // eslint-disable-next-line @typescript-eslint/no-empty-interface
   interface InverseAsymmetricMatchers extends Expect {}
 }
+
+declare namespace Vi {
+  // eslint-disable-next-line @typescript-eslint/no-empty-interface
+  interface AsymmetricMatchersContaining extends CustomMatchers {}
+}
```
rluvaton, line 886: Took the idea from

https://github.com/vitest-dev/vitest/blob/199776767a10bcd9f241e36a5cdd578c7d7c0dc2/test/core/test/jest-expect.test.ts#L9-L17

types/index.d.ts
```diff
@@ -1,5 +1,431 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
+interface CustomMatchers {
+  /**
+   * Note: Currently unimplemented
+   * Passing assertion
+   *
+   * @param {String} message
+   */
+  pass(message: string): void;
```
SimenB, line 10: why are these changed from `R` to `void`?

SimenB:
>@SimenB should I update the package version?

No need, I'll do that when publishing

types/index.d.ts
rluvaton: Weird, I just copy pasted

rluvaton: I'll do it again in a few minutes

rluvaton: Never mind, there shouldn't be any `R` I copy pasted from the Expect interface

Issue #518, opened 2022-10-16
Bad typing for `toHaveBeenCalledOnceWith` causing problems


Xunnamius:
**Bug**

- `package` version: 3.1.0
- `node` version: 18.10.0
- `npm` (or `yarn`) version: 8.19.2

What happened:

A test with the following line passes, but TypeScript complains with `Expected 0 arguments, but got 1. ts(2554)`:

```typescript
expect(revealSpy).toHaveBeenCalledOnceWith({ nodes: [node], index, parent });
```

`reveal` and `revealSpy` are defined as follows:

```typescript
const revealSpy = jest.spyOn(mdastUtilHidden, 'reveal').mockImplementation(() => undefined);

// Exported as part of mdastUtilHidden
function reveal<Nodes extends Hidden[]>({
  nodes,
  index,
  parent
}: {
  nodes: Nodes;
  index: number;
  parent: Parent<Node<Data>>;
}) {
  parent.children.splice(
    index,
    1,
    ...nodes.flatMap((n) => n.hiddenChildren.map((n) => removePosition(n)))
  );
}
```

I saw a recent commit described as "fixing toHaveBeenCalledOnceWith," so this may be a regression. Potentially related to #517. I've used `@ts-expect-error` to ignore the TS error for now. EDIT: `expect(...).toHaveBeenNthCalledWith(1, ...)` also works. Thanks for your work on this package!

Reproduction repository: https://github.com/Xunnamius/unified-utils/blob/b84a05ab0a10dd38b2c46062452f6c3e6cec194a/packages/mdast-util-hidden/test/unit-index.test.ts#L215

vandor:
I'm seeing the same problem.

I just started using `jest-extended`, specifically because I wanted `toHaveBeenCalledOnceWith`. It is offputting to get hints in my IDE like `Invalid number of arguments, expected 0` when using this matcher.

loyaj:
I'm seeing the same problem.

I just started using `jest-extended`, specifically because I wanted `toHaveBeenCalledOnceWith`. It is offputting to get hints in my IDE like `Invalid number of arguments, expected 0` when using this matcher.

I agree, it seems like this matcher was added recently and needs additional implementation

keeganwitt:
Yea, this was a simple goof in the type file. We currently lack test coverage of our types, but this is something we plan on adding.

konstantinmv:
Why has this been closed? Can we close the issue when this would be solved (with the new version, I suppose?)

[...]

PR #557, opened 2023-01-31
Support varargs in toHaveBeenCalledOnceWith (closes #517)

### LINKED ISSUE #517: toHaveBeenCalledOnceWith only asserts a match on the first parameter.
I'm not sure if this would be classed as a bug or a feature, so apologies for any mistakes raising the issue.

**Feature Request**

Description: toHaveBeenCalledOnceWith currently only matches against the first parameter provided to a mock function.

Possible solution:
In `toHaveBeenCalledOnceWith.js`, line 40 reads
`const pass = passOnce && this.equals(expected, received.mock.calls[0][0]);`
I believe that this should probably be changed to `this.equals(expected, received.mock.calls[0]);` and assert equality on the whole array of parameters passed to the mock function in the first call. This may need the function declaration to be changed to `toHaveBeenCalledOnceWith(received, ...expected)` or similar to handle multiple arguments.

If this is desired, or not so simple to implement, it would be good to make this clear in the documentation.

keeganwitt:


keeganwitt:
I'm not sure this is the right way to solve this. It doesn't seem right...

src/matchers/toHaveBeenCalledOnceWith.js
```diff
@@ -1,6 +1,6 @@
 import { isJestMockOrSpy } from '../utils';
 
-export function toHaveBeenCalledOnceWith(received, expected) {
+export function toHaveBeenCalledOnceWith(received, ...expected) {
```
Maxim-Mazurok, line 3: As [mentioned](https://github.com/jest-community/jest-extended/issues/518#issuecomment-1409835822) this PR should probably also change the types to match the new implementation

src/matchers/toHaveBeenCalledOnceWith.js
```diff
@@ -16,7 +16,9 @@ export function toHaveBeenCalledOnceWith(received, expected) {
   }
 
   const passOnce = received.mock.calls.length === 1;
-  const pass = passOnce && this.equals(expected, received.mock.calls[0][0]);
+  const mockCall = received.mock.calls[0];
+  const receivedArray = Array.isArray(mockCall) ? mockCall : [mockCall];
```
Maxim-Mazurok, line 20: I have checked `createToBeCalledWithMatcher` implementation in jest:
image

And I've tried using `mockCall` directly without converting it to array and it seems to pass all your tests:

image

Maxim-Mazurok (changes requested):

Maxim-Mazurok:
> I'm not sure this is the right way to solve this. It doesn't seem right...

Not sure what exactly seems not right, perhaps my second suggestion will make it seem right? Haven't touched much of jest internals tho, so I probably don't know better.

src/matchers/toHaveBeenCalledOnceWith.js
SimenB, line 20: It's always an array.

src/matchers/toHaveBeenCalledOnceWith.js
keeganwitt: Yea, I realized I missed that at first. Thank you. Fixed.

src/matchers/toHaveBeenCalledOnceWith.js
keeganwitt: Ah. I could have sworn I saw a case that wasn't, but I don't know what I thought I saw now 😛 
Removed the ternary.

Maxim-Mazurok:
Thanks so much! Looking forward to getting this published to npm, cheers!

PR #559, opened 2023-02-03
Fix exported type, resolve "is not a module" error in ts

### LINKED ISSUE #367: Typescript error when importing jest-extended

Codex-:
### What

Update the interface for matches to extend a record to provide the index string signature expected by `expect.extend`, matches what the `@testing-library` types do: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/testing-library__jest-dom/matchers.d.ts#L1

### Why

Fixes the error:
```
File '/node_modules/jest-extended/types/index.d.ts' is not a module.ts(2306)
```

Issue #367 was closed without addressing this problem, instead, the PR made that was used to close the issue describes a workaround. This workaround is not so suitable for other test suites consuming these matchers, like vitest without globals.

Fixes #367 

Currently, [this documentation](https://jest-extended.jestcommunity.dev/docs/getting-started/setup#use-with-vitest) errors with typescript, this fixes it and improves support.

### Notes

```ts
import * as matchers from 'jest-extended'
import { expect } from 'vitest'

matchers.fail // you can see all of the expected matchers provided by this package

expect.extend(matchers)

expect(1).pass("") // shows that using the existing type extensions, the new methods are still attached as expected
```

### Housekeeping

- [x] Unit tests

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
-interface CustomMatchers<R> {
+interface CustomMatchers<R> extends Record<string, any> {
   /**
    * Note: Currently unimplemented
    * Passing assertion
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -884,3 +884,8 @@ declare namespace Vi {
   // eslint-disable-next-line @typescript-eslint/no-empty-interface
   interface AsymmetricMatchersContaining extends CustomMatchers<any> {}
 }
+
+declare module 'jest-extended' {
+  const matchers: CustomMatchers<any>;
+  export = matchers;
+}


Codex-:
Hey @SimenB is there anything I can do to get some eyes on this? :)

Codex-:
@keeganwitt sorry to bother, would it be possible to have this looked at by yourself or a member of the team?

keeganwitt:
I'm not sure if there's anything else we need in the module you added. But this does seem to fix the issue. Thank you!

Codex-:
Thanks @keeganwitt , really appreciate it! :)

ooloth:
Any chance this change could be released soon? It would be handy.

keeganwitt:
We are working on a release now (#605).

PR #572, opened 2023-02-20
Fix toHaveBeenCalledOnceWith messages

keeganwitt:
The current messages print arrays, which is confusing when not invoked with an array. Also, I think the wording of the `.not` message could be clarified. Additionally, `${printReceived(received.mock.calls[0]?.[0])}` would only print the first argument of an invocation.


SimenB (approved):
works for me 👍 

remember to add a changeset if you want this to be published

PR #574, opened 2023-02-23
Rename toHaveBeenCalledOnceWith (closes #529)

### LINKED ISSUE #529: `toHaveBeenCalledOnceWith` naming is ambiguous
**Doc Update / Clarification and feature request for `toHaveBeenCalledNTimesWith`**

Just checked out the project as I was looking for a `toHaveBeenCalledNTimesWith` matcher, where `n` is most often 1. I think there's a bit of ambiguity around `toHaveBeenCalledOnceWith`.

- How it can be read: expect the mock function to have been called at most one time with specific arguments
- How it's [implemented](https://github.com/jest-community/jest-extended/blob/17bb7899857d047f78a528bbfd11ed7292f6c7a9/src/matchers/toHaveBeenCalledOnceWith.js#L40): expect the mock function to have been called only once, **and** when it was called, it should have been called with specific arguments.

In the former version - what I was after - the matcher doesn't care about how many times it was called in total, but it asserts that the mock should only have been called once with the arguments I provided.

It would be good to clarify this in the docks, and long-term a matcher like the former would be useful :)

keeganwitt:


smikheiev:
@keeganwitt do you know by any chance when will this be released?

keeganwitt:
We should cut a release soon I think. I'll try to get to that this week.

Issue #610, opened 2023-06-02
Vitest typings override instead of augmenting module


mlev:
**Bug**

The vitest typings still aren't quite right. The (currently unreleased) changes from PR https://github.com/jest-community/jest-extended/pull/600 are an improvement but still not working for me. I believe this is because the vitest module typings are being overwritten instead of augmented so you end up with errors like:

```
TS2305: Module '"vitest"' has no exported member 'expect'.
```

Also - it's currently only applying the matchers as Asymmetric but it should be updating the `Assertion` interface as well (as per vitest doco https://vitest.dev/guide/extending-matchers.html)

I can get it working by patching the typings locally - and changing the vitest declaration to this:

```
import "vitest";
declare module "vitest" {
  // eslint-disable-next-line @typescript-eslint/no-empty-interface
  interface Assertion<T = any> extends CustomMatchers<T> {}
  // eslint-disable-next-line @typescript-eslint/no-empty-interface
  interface AsymmetricMatchersContaining extends CustomMatchers<any> {}
}
```

I believe you need the top level import to make it an augmentation. However - this requires importing vitest and I'm not sure what will happen when using jest. It might be fine but not 100% sure.

An alternative would be not to include this declaration and let users do the configuration themselves. Not as nice - but better than current solution which overwrites the vitest types.

I'm using Typescript 5.0.4 with NodeNext/ESM enabled.

- `package` version: main branch
- `node` version: 18.6.0
- `npm` (or `yarn`) version: 9.5.1

burtek:
Just FFR - my workaround is to use code as below when I need `vitest` types rather than `jest-extend` types

```ts
import type { SpyInstance } from 'vitest/dist/index';
```

GuillaumeRahbari:
Sorry for the issue I introduced.

I made those 2 PRs
- https://github.com/jest-community/jest-extended/pull/613
- https://github.com/jest-community/jest-extended/pull/614

Both of them fixes the issue described by @mlev.

I do not have a perfect solution in mind. Each of the solution comes with advantages and drawbacks

GuillaumeRahbari:
In my side I was using jest-extended only for this functionality https://jest-extended.jestcommunity.dev/docs/matchers/Array#toincludesamemembersmembers

But I found out that vitest as this natively with the chai api. So I will probably not be using jest-extended anymore

Codex-:
Also running into this issue, for now having to drop the jest-extended version 😔

Might need some input from @keeganwitt

[...]

PR #600, opened 2023-05-10
feat: support typings for vitest 0.31.0

GuillaumeRahbari:
Feature
closes #598 

### What

Support of vitest 0.31.0. See the issue https://github.com/jest-community/jest-extended/issues/598

### Why

Vitest made a Breaking Change that can be seen [here](https://github.com/vitest-dev/vitest/releases/tag/v0.31.0)

### Housekeeping

- [ ] Unit tests
- [ ] Documentation is up to date
- [ ] No additional lint warnings
- [x] [Typescript definitions](https://github.com/jest-community/jest-extended/blob/main/types/index.d.ts) are added/updated where relevant

Issue #598, opened 2023-05-04
Breaking change in vitest


GuillaumeRahbari:
**Feature Request**

Description: 

Vitest just released the version [0.31.0](https://github.com/vitest-dev/vitest/releases/tag/v0.31.0). They changed the way to declare module in typescript.

Possible solution:

Add in the file https://github.com/jest-community/jest-extended/blob/main/types/index.d.ts the following code to be compatible with every version
```ts
declare module 'vitest' {
  // eslint-disable-next-line @typescript-eslint/no-empty-interface
  interface AsymmetricMatchersContaining extends CustomMatchers<any> {}
}
```

PR #614, opened 2023-06-09
fix: support vitest 0.31.0

GuillaumeRahbari:
Bug
closes https://github.com/jest-community/jest-extended/issues/610

### What

Support of vitest 0.31.0. See the issue https://github.com/jest-community/jest-extended/issues/598

### Why

Vitest made a Breaking Change that can be seen [here](https://github.com/vitest-dev/vitest/releases/tag/v0.31.0)
This [PR](https://github.com/jest-community/jest-extended/pull/600) did not allow to fully support vitest types

### Housekeeping

- [ ] Unit tests

### DIFF
--- a/.changeset/itchy-chicken-admire.md
+++ b/.changeset/itchy-chicken-admire.md
@@ -1,0 +1,5 @@
+---
+'jest-extended': patch
+---
+
+fix compatibility with vitest 0.31.0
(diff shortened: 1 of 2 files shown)


oskarols:
Wouldn't it be better to have the vitest-specific import in e.g. `jest-extended/vitest`? This way the Jest users wouldn't be bothered by the type import.

keeganwitt:
I think for the pipeline to pass, we'd need a dependency on vitest. I don't think we'd want to do that.

LukeNotable:
How about supplying an empty type stub for vitest just for lint/typecheck?

keeganwitt:
I'm going to close this PR for now, but please feel welcome to open another one if there are ideas for alternative approaches we should discuss.

PR #636, opened 2023-07-25
fix: vitest types not working (closes #610)

### LINKED ISSUE #610: Vitest typings override instead of augmenting module

keeganwitt:


SimenB:
(don't forget the changeset)

website/docs/getting-started/setup.md
```diff
@@ -52,7 +51,47 @@ Add this setup script to your `vitest.config.js`:
...
   test: {
-    setupFiles: ["./testSetup.js"],
+    setupFiles: ['./testSetup.js'],
   },
 });
 ```
+
+### Vitest TypeScript types setup
+
+To use Vitest with TypeScript, create a file named _jest-extended.d.ts_ with the content below in addition to the setup above.
```
SimenB, line 61:
```suggestion
To use Vitest with TypeScript, create a file named `jest-extended.d.ts` with the content below in addition to the setup above.
```

PR #643, opened 2023-08-07
Add matcher .toHaveSameJson

giancosta86:
The purpose of this pull requests consists in introducing the `.toHaveSameJson` matcher - which ensures that the `actual` value has the same JSON serialization as the `expected` value.


giancosta86:
TypeScript version added to [more-jest](https://github.com/giancosta86/more-jest).

PR #647, opened 2023-09-05
docs: clarify the difference between toIncludeAllMembers and toIncludeSameMembers

### LINKED ISSUE #609: Better clarify the difference between toIncludeAllMembers and toIncludeSameMembers in the documentation
**Feature Request**
Description:
Only by reading the names is not very straightforward to reason about the difference immediately, specially when the description of `toIncludeAllMembers` mentions `the same members`.

The difference as I understand is this:

```ts
expect([1, 2, 3]).toIncludeSameMembers([1, 2, 3]); // passes, as array elements are equal, extra elements are not allowed
expect([1, 2, 3, 4]).toIncludeAllMembers([1, 2, 3]); // passes, even though there's an extra 4, it also contains [1, 2, 3]
```

Possible solution:
- Improve the description text, emphasizing that for `toIncludeSameMembers` no extra values are allowed.
- Add negative test cases (e.g using .not) instead of just positive examples.

devjiwonchoi:
### What

This PR updates the description of `toIncludeAllPartialMembers` by removing the `'same'` keyword.
Adds `not` test example to the docs.
Also, updates the description in the types, and adds a test code corresponding to the docs.

### Why

It confuses the reader with `toIncludeSameMembers` and `toIncludeAllMembers`

### Notes

Fixes #609 

### Housekeeping

- [x] Unit tests

### DIFF
--- a/website/docs/matchers/Array.mdx
+++ b/website/docs/matchers/Array.mdx
@@ -28,18 +28,19 @@ Use `.toBeArrayOfSize` when checking if a value is an `Array` of size x.
 
 ### .toIncludeAllMembers([members])
 
-Use `.toIncludeAllMembers` when checking if an `Array` contains all of the same members of a given set.
+Use `.toIncludeAllMembers` when checking if an `Array` contains all of the members of a given set.
 
 <TestFile name="toIncludeAllMembers">
   {`test('passes when given array values match the members of the set', () => {
   expect([1, 2, 3]).toIncludeAllMembers([2, 1, 3]);
   expect([1, 2, 2]).toIncludeAllMembers([2, 1]);
+  expect([1, 2, 2]).not.toIncludeAllMembers([3, 1]);
 });`}
 </TestFile>
 
 ### .toIncludeAllPartialMembers([members])
 
-Use `.toIncludeAllPartialMembers` when checking if an `Array` contains all of the same partial members of a given set.
+Use `.toIncludeAllPartialMembers` when checking if an `Array` contains all of the partial members of a given set.
 
 <TestFile name="toIncludeAllPartialMembers">
   {`test('passes when given array values match the partial members of the set', () => {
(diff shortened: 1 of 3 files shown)


keeganwitt:
Thanks for the PR!

I can certainly understand why these two matchers would be confused. I had to read it a few times before I understood myself. But I'm not sure the removal of the word "same" makes it any clearer.

I think this makes it clearest

```javascript
expect([1, 2, 2]).toIncludeAllMembers([2, 1]);  // passes
expect([1, 2, 2]).toIncludeSameMembers([2, 1]);  // fails
```

* `toIncludeAllMembers` passes when all members of the test set are in the array (any order).
* `toIncludeSameMembers` passes when the arrays match (any order).

I'm for improving the documentation, but I think probably a more extensive edit is needed to improve on what we have.

devjiwonchoi:
@keeganwitt I'll update the docs, thank you!

For `toIncludeAllPartialMembers`, I don't understand the mechanism.

This test case passes, and I'm not sure why. Could it be a bug?

```js
expect([1, 2, 2]).toIncludeAllPartialMembers([999]);
```

website/docs/matchers/Array.mdx
```diff
@@ -62,13 +62,13 @@ Use `.toIncludeAnyMembers` when checking if an `Array` contains any of the membe
...
 
-Use `.toIncludeSameMembers` when checking if two arrays contain equal values, in any order.
+Use `.toIncludeSameMembers` when checking if two arrays contain members in exact match, in any order.
 
 <TestFile name="toIncludeSameMembers">
   {`test('passes when arrays match in a different order', () => {
   expect([1, 2, 3]).toIncludeSameMembers([3, 1, 2]);
   expect([{ foo: 'bar' }, { baz: 'qux' }]).toIncludeSameMembers([{ baz: 'qux' }, { foo: 'bar' }]);
-  expect([{ foo: 'bar' }, { baz: 'qux' }, { fred: 'thud' }]).not.toIncludeSameMembers([{ baz: 'qux' }, { foo: 'bar' }]);
+  expect([1, 2, 2]).not.toIncludeSameMembers([2, 1]);
```
keeganwitt, line 71: Why change this?

website/docs/matchers/Array.mdx
```diff
@@ -62,13 +62,13 @@ Use `.toIncludeAnyMembers` when checking if an `Array` contains any of the membe
 
 ### .toIncludeSameMembers([members])
 
-Use `.toIncludeSameMembers` when checking if two arrays contain equal values, in any order.
+Use `.toIncludeSameMembers` when checking if two arrays contain members in exact match, in any order.
```
keeganwitt, line 65: I'm not sure this is clearer. The matcher is using the `equals` function, which is why the original was worded the way it was.

keeganwitt:
> ```js
> expect([1, 2, 2]).toIncludeAllPartialMembers([999]);
> ```

It's because the matcher is expecting to compare objects. That's what is meant by "partial" is that some of the keys in the object match. I wonder if this matcher should be failing if the type isn't an object 🤔

PR #650, opened 2023-09-11
Mark toHaveBeenCalledBefore and After's second parameter as optional

Blond11516:
### What

Marks `toHaveBeenCalledBefore` and `toHaveBeenCalledAfter`'s second parameter as optional in the type declarations.

### Why

Current type declarations generate an error if the parameter isn't passed. This is inaccurate, as the default value will be used in this case.

### Notes

Fixes https://github.com/jest-community/jest-extended/issues/651 

### Housekeeping

- [ ] Unit tests

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -156,7 +156,7 @@ interface CustomMatchers<R> extends Record<string, any> {
    * @param {Mock} mock
    * @param {boolean} [failIfNoFirstInvocation=true]
    */
-  toHaveBeenCalledAfter(mock: jest.MockInstance<any, any[]>, failIfNoFirstInvocation: boolean): R;
+  toHaveBeenCalledAfter(mock: jest.MockInstance<any, any[]>, failIfNoFirstInvocation?: boolean): R;
 
   /**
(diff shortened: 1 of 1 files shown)


chornos13:
I think you forgot to mark the second argument of toHaveBeenCalledAfter function as an optional

Blond11516:
@chornos13 I didn't personally need `toHaveBeenCalledAfter` when I opened the PR but I think it makes sense to do both at the same time. I've updated the PR.

keeganwitt:
Thanks for the PR!

SimenB:
Should we add an action that checks if either there's an added changeset or we add a label? Seems the message from the bots is overlooked by contributors most of the time 😅

keeganwitt:
> Not all PRs need a changeset - e.g. docs only. But seeing as the comment is ignored, we might need something that yells louder before merging

Oh, I thought docs used the changeset for publishing the updated docs to the website.

SimenB:
Nah, changesets is solely for publishing to npm.

Website changes are apparently a manual step (https://github.com/jest-community/jest-extended/pull/648#issuecomment-1727195061) - ideally that would just be autopublished on pushes to `main`. That's how https://jestjs.io/ works

Issue #652, opened 2023-09-22
Declaration merging does not work without exporting the CustomMatchers interface.


cdierkens:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `jest-extended@4.0.1` for the project I'm working on.

In order for declaration merging to work, the `CustomMatchers` interface needs to be exported.

**Before**
<img width="489" alt="image" src="https://github.com/jest-community/jest-extended/assets/4812409/137eca62-23fd-41e2-ab59-e332c0be5894">

**After**
<img width="483" alt="image" src="https://github.com/jest-community/jest-extended/assets/4812409/2f8a012e-2423-40cd-9cad-c16754c8d888">

**Full .d.ts file**

```ts
import type { CustomMatchers } from "jest-extended";
import "vitest";

declare module "vitest" {
  interface Assertion<T = any> extends CustomMatchers<T> {}
  interface AsymmetricMatchersContaining<T = any> extends CustomMatchers<T> {}
}
```

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/jest-extended/types/index.d.ts b/node_modules/jest-extended/types/index.d.ts
index a0cef96..bc8722a 100644
--- a/node_modules/jest-extended/types/index.d.ts
+++ b/node_modules/jest-extended/types/index.d.ts
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
-interface CustomMatchers<R> extends Record<string, any> {
+export interface CustomMatchers<R> extends Record<string, any> {
   /**
    * Note: Currently unimplemented
    * Passing assertion
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

cdierkens:
Closing this issue.

I was able to get declaration merging working with the following `jest-extended.d.ts` file.

```ts
import type CustomMatchers from "jest-extended";
import "vitest";

declare module "vitest" {
  interface Assertion<T = any> extends CustomMatchers<T> {}
  interface AsymmetricMatchersContaining<T = any> extends CustomMatchers<T> {}
  interface ExpectStatic extends CustomMatchers<T> {}
}
```

charliematters:
Sorry to dig up a closed issue, but this looks like invalid TS to me?
``` typescript
 interface ExpectStatic extends CustomMatchers<T> {}
```
[...]

PR #653, opened 2023-09-22
fix: export CustomMatchers interface for declartion merging

cdierkens:
### What

export CustomMatchers interface for declartion merging

### Why

https://github.com/jest-community/jest-extended/issues/652

### Housekeeping

- [x] Unit tests

### DIFF
--- a/.changeset/beige-fans-tap.md
+++ b/.changeset/beige-fans-tap.md
@@ -1,0 +1,5 @@
+---
+"jest-extended": patch
+---
+
+fix: export CustomMatchers interface for declartion merging
--- a/.changeset/beige-fans-tap.md
+++ b/.changeset/beige-fans-tap.md
@@ -2,4 +2,4 @@
 "jest-extended": patch
 ---
 
-fix: export CustomMatchers interface for declartion merging
+Export `CustomMatchers` interface for declartion merging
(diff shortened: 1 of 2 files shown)


SimenB (approved):
thanks!

SimenB:
Huh, unsure why this change breaks the typecheck 🤔

cdierkens:
> Huh, unsure why this change breaks the typecheck 🤔

@SimenB I'll try updating the branch. Maybe there's something else I can look into if that doesn't work?

cdierkens:
I'll run the type checking locally and do a bit more research on how other libraries go about declaration merging.

keeganwitt:
I tried adding a dependency on `@types/jest` and adding `/// <reference types="jest" />`, but that didn't help.

keeganwitt:
I'm not sure if this is the right way to fix this, but this way passes the check.

@cdierkens can you confirm whether this still works as a patch for you?

cdierkens:
@keeganwitt I was able to get declaration merging working with the existing types!

```ts
import type CustomMatchers from "jest-extended";
import "vitest";

declare module "vitest" {
  interface Assertion<T = any> extends CustomMatchers<T> {}
  interface AsymmetricMatchersContaining<T = any> extends CustomMatchers<T> {}
  interface ExpectStatic extends CustomMatchers<T> {}
}

```

PR #664, opened 2023-10-09
Allow `fail` to work in `try/catch`

### LINKED ISSUE #663: `fail` does not work when testing code using a `try/catch`
**Bug**

- `package` version: 4.0.1
- `node` version: 18.18.0
- `yarn` version: 1.22.19

Relevant code or config

```javascript
// my.test.js
function invokeWithTry(callback) {
  try {
    return callback();
  } catch (err) {
    return err;
  }
}


test("fail works in a try/catch", () => {
  const result = invokeWithTry(() => expect().fail("expected failure")); // Erroneously passes
  expect(result).toBeUndefined(); // Actually fails here b/c `result` is a `JestAssertionError`
});
```
```jsonc
// package.json
{
  "devDependencies": {
    "jest": "^29.7.0",
    "jest-extended": "^4.0.1"
  },
  "jest" : {
    "setupFilesAfterEnv": ["jest-extended/all"]
  }
}
```

**What you did:**

I passed a callback containing `expect().fail()` to a function that invokes the callback in a `try/catch` block.

**What happened (please provide anything you think will help):**

The failure was not registered since the `catch` intercepted the `JestAssertionError` raised by the `fail` matcher.

Reproduction repository (if possible): https://github.com/thehale/jest-extended-fail-not-working-in-try-catch

thehale:
### What

This PR allows calls to `expect().fail(message)` to fail a test, even when the `expect().fail()` is executed inside of a try/catch block.

### Why

closes #663 

This functionality has been [desired in Jest](https://github.com/jestjs/jest/issues/11698#issuecomment-1135857402) since the original `fail` function was removed from the imported Jasmine scope.

Throwing an error is semantically different from failing a test, and attempting to equate the two makes it much harder to test code which uses try/catch blocks.

### Notes

I am unaware of any way for Jest to report an "expected failure" similar to `pytest`s [`xfail` decorator](https://docs.pytest.org/en/7.3.x/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail). The closest I could find was prefixing the `test` as `xtest` to skip it. Deleting those `x` prefixes allows one to manually inspect the test failure messages to verify they line up with the expected behavior.

I am happy to edit the PR if there's a better way to record "expected failures".

### Housekeeping

- [x] Unit tests

### DIFF
--- a/src/matchers/fail.js
+++ b/src/matchers/fail.js
@@ -1,4 +1,5 @@
 export function fail(_, message) {
+  this.dontThrow();
   return {
     pass: false,
     message: () => (message ? message : 'fails by .fail() assertion'),
--- a/test/matchers/__snapshots__/fail.test.js.snap
+++ b/test/matchers/__snapshots__/fail.test.js.snap
@@ -1,5 +1,0 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
-
-exports[`.fail fails with message 1`] = `"This shouldn't fail!"`;
-
-exports[`.fail fails without message 1`] = `"fails by .fail() assertion"`;
--- a/test/matchers/fail.test.js
+++ b/test/matchers/fail.test.js
@@ -3,11 +3,18 @@ import * as matcher from 'src/matchers/fail';
 expect.extend(matcher);
 
 describe('.fail', () => {
-  test('fails without message', () => {
-    expect(() => expect().fail()).toThrowErrorMatchingSnapshot();
+  xtest('fails without message', () => {
+    expect().fail(); // This should fail!
   });
-  test('fails with message', () => {
-    expect(() => expect().fail("This shouldn't fail!")).toThrowErrorMatchingSnapshot();
+  xtest('fails with message', () => {
+    expect().fail('This should fail!');
+  });
+  xtest('fails when invoked in a try/catch', () => {
+    try {
+      expect().fail();
+    } catch (error) {
+      expect('this assertion').toBe('not checked');
+    }
   });
 });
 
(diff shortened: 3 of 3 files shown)


SimenB:
Having skipped tests doesn't make much sense. We probably need some sort of integration test which spawns Jest or something for this case

liamjones:
I'm also looking for this functionality but I don't think this solution works as-is.

`xtest` aka `test.skip` is, as SimenB said, is skipping the test.

I think what you wanted instead was [`test.failing`](https://jestjs.io/docs/api#testfailingname-fn-timeout) which is considered good when the test fails and bad when it passes.

If I checkout your PR and change the `xtest` to `test.failing` the test, incorrectly, passes.

thehale:
@liamjones The docs seem to suggest that `test.failing` only flips the expectation if the test **throws** and error. 

> If failing test will throw any errors then it will pass. If it does not throw it will fail.
> > https://jestjs.io/docs/api#testfailingname-fn-timeout

Since this change directly reports a test as failing *instead* of throwing an exception for Jest to catch, `test.failing` never catches a `JestException`, causing it to mistakenly believe the test passed.

Notice that if you change `xtest` to `test`, the tests do fail as expected.

thehale:
After digging into all this some more, I figured out a way to use a [custom jest reporter](https://jestjs.io/docs/configuration#custom-reporters) to detect the expected test failures and flip them to passing. 

The custom reporter also marks these "expected failing" tests as actually failing if they unexpectedly pass.

The solution is decently complex, so I'm happy to answer questions about how it works and implement suggestions to make the implementation more approachable.

thehale:
@SimenB This PR is ready for another maintainer review.

*P.S. I see per your GitHub status that you are currently "busy"/"slow to respond", so no rush!*

PR #687, opened 2024-05-20
Upgrade Yarn off of classic

keeganwitt:
### What

Upgrade off of classic Yarn to the latest Yarn.

### Why

This makes jest-extended match Jest and provides the [benefits](https://yarnpkg.com/migration/overview) Yarn documents.


keeganwitt:
This will require Node 18 or newer. Otherwise you get an error like below.

```
/usr/local/bin/yarn --version
Usage Error: This tool requires a Node version compatible with >=18.12.0 (got 16.20.2). Upgrade Node, or set `YARN_IGNORE_NODE=1` in your environment.
```

keeganwitt:
> This will require Node 18 or newer. Otherwise you get an error like below.
> 
> ```
> /usr/local/bin/yarn --version
> Usage Error: This tool requires a Node version compatible with >=18.12.0 (got 16.20.2). Upgrade Node, or set `YARN_IGNORE_NODE=1` in your environment.
> ```

Simen pointed out that this was a limitation of Yarn 4 and we can retain the versions that were currently have with Yarn 3, which Jest is using.

.gitignore
```diff
@@ -8,3 +8,5 @@ coverage
 .vs
 .vscode
 package-lock.json
+/.yarn/install-state.gz
+**/.pnp.*
```
SimenB, line 12: see https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored

keeganwitt: Added their recommendations.

PR #688, opened 2024-05-22
Update Node.js supported engines and CI test versions

keeganwitt:
Add Node 20 and 22 to supported engines, and add 21 and 22 to CI test versions.


keeganwitt:
We could wait a week then update this PR since 22 is about to go LTS.

PR #689, opened 2024-05-26
feat: add toIncludeCaseInsensitive matcher

mts88:
Add the new String matcher .toIncludeCaseInsensitive

### Housekeeping

- [ ] Unit tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ coverage
 .vs
 .vscode
 package-lock.json
+yarn.lock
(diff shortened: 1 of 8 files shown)


keeganwitt:
I wonder if it'd be better to add an optional argument to the `toInclude` matcher to make it case insensitive instead of adding a new matcher. What do you think?

mts88:
> I wonder if it'd be better to add an optional argument to the `toInclude` matcher to make it case insensitive instead of adding a new matcher. What do you think?

Interesting suggestion! I can edit it and introduce the case-insensitive option to the `toInclude` matcher

PR #693, opened 2024-07-04
feat: Add actual and expected fields to the return value of toBeOneOf

hzmming:
Add actual and expected fields to the return value of toBeOneOf

When using the testResultsProcessor of jest.config.js, I need to obtain the actual and expected values for processing and display

### Housekeeping

- [x] Unit tests

### DIFF
--- a/.changeset/proud-cameras-pull.md
+++ b/.changeset/proud-cameras-pull.md
@@ -1,0 +1,5 @@
+---
+'jest-extended': patch
+---
+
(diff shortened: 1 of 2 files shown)


keeganwitt:
The return values are defined by the [custom matchers API](https://jestjs.io/docs/expect#custom-matchers-api), which only has two keys: `pass` and `message`. And the message already includes the actual and expected.

PR #695, opened 2024-07-19
rm redundant strict... compiler options in tsconfig.json

dandv:
All these properties are true already because `strict` was set to true.


keeganwitt:
Thanks for the PR!

PR #743, opened 2025-04-15
Typescript support

keeganwitt:
~~Uses tsc for building, as a first step to moving the library to TypeScript.~~

I had planned on leaving all the files as js instead of ts. But this didn't work with ts-jest with Jest 27 and 28 (but did with 29). So I renamed the files also and removed js file processing from ts-jest and tsc.


keeganwitt:
I had planned on leaving all the files as js instead of ts. But this didn't work with ts-jest with Jest 27 and 28 (but did with 29). So I renamed the files also and removed js file processing from ts-jest and tsc.

PR #772, opened 2025-05-19
feat: Add BigInt support to number matchers (closes #591)

### LINKED ISSUE #591: BigInt support for Number functions
**Feature Request**

Description: For each Number related matcher, support BigInt input values

For example this test should pass:

```ts
test('passes when value is a number', () => {
  expect(1n).toBeNumber()
});
```

This matches Jest behaviour: https://github.com/facebook/jest/pull/8382 

```ts
test('passes when value is a number', () => {
  expect(1n).toBeGreaterThan(0)
});
```

Possible solution: Probably need to update each matcher on a case-by-case basis. Initial thoughts for each, might not cover everything: 

- `toBeNumber` is just a case of updating the typeof check
- `toBeNaN` should be able to be fixed by changing from `isNaN` to `Number.isNaN` which supports bigint correctly
- `isFinite` will probably need to cast its input to `Number` since `Number.isFinite(123n)` is false for some reason
- `toBePositive` uses `isNaN`, probably just needs updating to `Number.isNaN` 🤞 
- `toBeNegative` seems to work (should add tests though to make sure it keeps working!)
- `toBeEven` probably needs to check the type, and use `2n` instead of `2` for bigint (can't do bigint % number, or vice versa)
- `toBeOdd` same as `toBeEven`
- `toBeWithin` seems to work
- `toBeInteger` will probably need to cast its input to `Number` since `Number.isInteger(123n)` is false for some reason

keeganwitt:


keeganwitt:
Actually, since we have `toBeInteger`, it might make more sense to add `toBeBigInt` instead of changing the behavior of this function.

PR #805, opened 2025-08-01
Update setup.md

Yanis-Kouidri:
Add instruction to easily setup jest-extended with a test.config.mjs file. That what I was looking for when I wanted to setup jest-extended

### What

I updated the documentation , I add a paragraph in the setup part

### Why

Because when I followed the current setup instruction I was confused. Just adding one line in test.config.mjs is enough to setup jest-extended feature.

### Housekeeping

- [x] Unit tests


keeganwitt:
Thanks for the PR!

